PR #1, opened 2019-03-07
parse sentences

Gregoor:
TODOS:
- [ ] put end-of-sentence punctuation back in
- [x] stop filtering for invalid characters for now


Gregoor:
@fiji-flo is there something like `str.split` but where you split after the given pattern (so that we can keep the punctuation)? I've tried writing something using `vec.fold` but that turned into a bit of an unidiomatic mess.

fiji-flo:
Not to my knowledge. It guess we could do this either with `matches` or implement a custom split.

PR #2, opened 2019-03-09
SentenceExtractor iter

fiji-flo:
Didn't fix the . Let's see if we want to.

### DIFF
--- a/src/extractor.rs
+++ b/src/extractor.rs
@@ -1,0 +1,60 @@
+use std::iter::Iterator;
+use std::iter::Peekable;
+use std::vec::IntoIter;
+
+static FULL_STOP: char = '。';
+
+/// An Iterator extracting _sentences_ from a text node.
+/// We throw away:
+/// - Anything be fore the first [FULL_STOP]
+/// - Anthing where [contains_no_invalid_char] returns `false`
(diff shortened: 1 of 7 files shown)


src/extractor.rs
```diff
@@ -0,0 +1,60 @@
+use std::iter::Iterator;
+use std::iter::Peekable;
+use std::vec::IntoIter;
+
+static FULL_STOP: char = '。';
```
Gregoor: we'd probably want all the punctuation `'。', '？', '！'`

src/extractor.rs
```diff
@@ -0,0 +1,60 @@
...
+        SentenceExtractor {
+            sentences: extract_sentences(sentences),
+        }
+    }
+}
+
+impl<'a> Iterator for SentenceExtractor<'a> {
+    type Item = &'a str;
+    fn next(&mut self) -> Option<&'a str> {
+        if self.sentences.peek().is_some() {
```
Gregoor: why not return `self.sentences.next()` directly? i.e. wouldn't the API of this function be the same without the conditional?

Gregoor:
Apparently we want to keep the punctuation in, shall I try to tackle this?

PR #20, opened 2019-06-27
Add generic checker with config (fixes #16)

### LINKED ISSUE #16: Implement config for locales
Currently we have modules for each language. To make sure that contributors can come up with their own rules and test them, we should generalize it.

* Extractor reads rules from a config file
* There is one module to handle all the config cases
* Detection is automatic depending on the language that is passed to the script

MichaelKohler:


MichaelKohler:
@fiji-flo r?

PR #37, opened 2019-07-16
Add config option for disallowed words

Gregoor:


Gregoor:
Oh yeah, just looked at the spanish file. Guess a separate file makes sense. So I'll just remove the config key again and extend `load_config` to optionally look for a `<lang>_disallowed_words.txt` file with words separated by line breaks?

MichaelKohler:
I'd go for a separate folder, so disallowed_words/spanish.txt for example. :)

nukeador:
Let me know when this is ready and I can test it with the extracted dump and blacklist file we already have for Spanish (while we work on getting the blacklist for German too)

Gregoor:
Separate file it is. Config key is still also in the config, cause I didn't find a neat way to have a struct with an additional field from another (also the `..`-struct-update apparently only works for structs of the same type). Thanks for the feedback!

nukeador:
I'm not able to see how to add the blacklist file path to the config

I tried 

```disallowed_words = ["src/rules/disallowed_words/spanish.txt"]```

and

disallowed_words = ["./src/rules/disallowed_words/spanish.txt"]

But it's including words in the blacklist.

**Comment by Gregoor** [comment]
Oh sorry that was misleading. You can ignore the spanish.toml config key (unless you only want to specify a few words). In the rules dir there's now also a dir called `disallowed_words`. If you create a file called `spanish.txt` it should just work

**Comment by nukeador** [comment]
It seems that if a blacklisted word is followed by a comma or fullstop it's not filtered.

**Comment by Gregoor** [comment]
Good point, fixed!

**Comment by nukeador** [comment]
I also see that if a word is lower case on the blacklist and then appears as first letter uppercase, it's not filtered.

Stenotaphrum vs stenotaphrum

They should be the same word and filtered out.

**Comment by Gregoor** [comment]
That shouldn't happen, I lowercased both the list of disallowed words and each word that is being checked. Maybe there's some other characters behind stenotaphrum on the list? I don't trim anything from those lines, maybe I should.

**Comment by Gregoor** [comment]
Alright, can you give it another shot with the most recent version @nukeador? Now everything non-alphabetic will be trimmed.

**Comment by nukeador** [comment]
blacklist if failing if the word is inside double quotes ""

**Comment by nukeador** [comment]
Uppercase first letter is still failing

**Comment by nukeador** [comment]
I think right now the file is defined to be toml, not a txt

https://github.com/Common-Voice/common-voice-wiki-scraper/blob/38510a021d0925449657a9703327c0070259a9f1/src/config.rs#L18

Changing the txt to toml, it loads the blacklist.

**Comment by Gregoor** [comment]
Oh dang, my bad. I thought I had a test case that covered the list, but apparently some other rule kicked in 🙈 Sorry for the confusion and thanks for catching that. I'll change it to txt!

**Comment by nukeador** [comment]
With the latest changes I see words in the blacklist being scrapped. Note I'm using the file and also individual words on the config.

**Comment by nukeador** [comment]
My bad, I didn't realized we fixed the .txt issue, renaming blacklist files now seems to be filtering.

**Comment by nukeador** [comment]
@Gregoor I've added a commit with the Spanish and German blacklists, it should be ready for merge.

File: README.md [hunk 1]
```diff
@@ -55,4 +56,46 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_letter_start |  If a sentence needs to start with a letter | boolean | true
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 
+## Using disallowed words (blacklisting)
 
+In order to increase the quality of the final output, you might want to consider filtering out some words that are complex, too long or non-native.
+
+You can do this by adding these words to the language rules file for your language under the disallowed_words setting.
+
+If your list of too long, you can also place a `<language>.txt` file in the `rules/disallowed_words` directory to enable a new locale. Each word should be in a new line.
```
MichaelKohler: -> "If your list of disallowed words is too long"?

nukeador:
@Gregoor Do you think we can merge this now so I can share a post this week to the more technical CV contributors to start playing around? I think it would be good to help us get a point this is more automated and doesn't require a technical person to get sentences samples.

Gregoor:
@nukeador Yes, fair enough, let's merge it and track the long-blocklist issue separately.

PR #38, opened 2019-07-29
Update README.md: fix some code example

jaumeortola:


README.md
```diff
@@ -74,7 +74,7 @@ After running step 1 and 2 from the Usage section above, run:
 
 ```bash
 cd ../common-voice-wiki-scraper
-cargo run -- extract ../wikiextractor/text/ --no-check >> wiki.en.all.txt
+cargo run -- extract -d ../wikiextractor/text/ --no_check >> wiki.en.all.txt
```
MichaelKohler: I think this could also include the -l tag to make it more obvious that a language is needed, as this most probably won't be run against English rules?

jaumeortola: It's a "full export with all wikipedia sentences". No rules are applied here.

MichaelKohler:
Oh, you're absolutely right. Nevermind me :)

PR #47, opened 2019-08-29
Esperanto rule file and Blacklist

stefangrotz:
Hello, unfortunately I also committed the sentences, please exclude these files from the commit, I don't have the rights to do this.


nukeador:
Can you please remove the files and squash the commits?

We would also details about the output of these rules:

- How many sentences are you getting?
- How did you create the blacklist?
- Get 2-3 additional native speakers (ideally some linguistics) to comment here with the estimated error rate.

Thanks!

stefangrotz:
Thanks for the feedback, I will create another pull request with the information in a few days, I just got a few ideas what things I should improve in the rules.

PR #52, opened 2019-10-10
Use punkt from GitHub for now

MichaelKohler:
Punkt 1.0.5 does not work with latest Rust nightlies. The issue was fixed in the master branch, however there hasn't been a release of it. Therefore I'm changing the dependency to use the git for now. Not the best practice, I know..


MichaelKohler:
@fiji-flo r?

MichaelKohler:
I'll test again if that is still the case once I run the esperanto export (yep, gonna do that at some point)...

MichaelKohler:
I'm closing this, as I think it's not necessary anymore.

PR #53, opened 2019-10-12
Add config to check for uneven symbols (fixes #32)

### LINKED ISSUE #32: Unclosed " or double empty ""
I am testing the italian rules and I see a lot of sentences like:
```
L'anno seguente fu acquistata dalla "Commodore International Ltd.
Esso non è un bene per nessuno"".
Nemini nimium bene est significa letteralmente: ""perché aspiriamo al superfluo?
```

Those are broken sentences because are unclosed or there are empty "", I don't think that is a regex problem in the rules file but something that the scraper consider as good sentences but they aren't and probably also other languages can have the same problem.

MichaelKohler:


MichaelKohler:
@fiji-flo r?

MichaelKohler:
@nukeador r?

PR #55, opened 2019-11-22
Italian rules updated

Mte90:
I am working to improve the blacklist but I saw this parameters that can be improved.

### DIFF
--- a/src/rules/italian.toml
+++ b/src/rules/italian.toml
@@ -1,7 +1,7 @@
 min_trimmed_length = 3
 min_word_count = 2
-max_word_count = 14
-min_characters = 0
+max_word_count = 16
+min_characters = 2
 may_end_with_colon = false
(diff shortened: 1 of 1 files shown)


MichaelKohler:
The broken whitespace and min characters changes look good to me. What data is the maximum words based on? Would love to hear more about that :)

Mte90:
Italian has sentences with more words then other languages, I saw also that sometimes are shorts so it was a workaround for that.

MichaelKohler:
@nukeador any thoughts on the word count increase? The rest looks good to me.

nukeador:
As long as the resulting sentences take less than 10s to read, we should be fine.

MichaelKohler:
@Mte90 can you confirm? If so, I'll merge

Mte90:
Yes I can confirm :-)

PR #58, opened 2019-12-10
Add continuous integration (AppVeyor & Travis CI)

EwoutH:
Adds continuous integration with both AppVeyor and Travis CI. 

For optimal integration with GitHub, [AppVeyor](https://github.com/marketplace/appveyor) and [Travis CI](https://github.com/marketplace/travis-ci) needs to be enabled on the GitHub Marketplace.

AppVeyor builds & tests on Windows with Rust nightly, while Travis builds & tests on Linux with Rust nightly.

AppVeyor also publishes the artifacts with each build. When tagging a new (pre)release, AppVeyor will also upload `libcommon_voice_yotp.rlib`, `libcommon_voice_yotp.d`, `common_voice_yotp.pdb`, `common_voice_yotp.exe` and `common_voice_yotp.d` to GitHub, creating a release page [like this](https://github.com/EwoutH/common-voice-wiki-scraper/releases/tag/2019-12-10-test0).

To enable uploading to GitHub Releases, you need update this patch with your personal authentication token. You can generate one on https://github.com/settings/tokens/new (select `public_repo` as scope), and then encrypt it with https://ci.appveyor.com/tools/encrypt. Replace the current token on line 34 (after `secure:`) in the `appveyor.yml` file.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,9 +19,8 @@ install:
 build_script:
     - cargo build --release --target=%target%      # Builds file defined in Cargo.toml (default main.rs)
 
-#test_script:
-#    - cargo test --target=%target% --verbose       # Runs tests in "src" and "tests" folders
-#    - cargo bench --target=%target% --verbose      # Runs benchmarks in "src/lib.rs"
+test_script:
+    - cargo test --target=%target% --verbose       # Runs tests in "src" and "tests" folders
 
 artifacts:
     - path: target\$(target)\release\*voice*.*     # Publishes all files from `cargo build/test/bench --release`
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,10 +10,10 @@ environment:
 install:
     - git submodule update --init
     - appveyor DownloadFile https://win.rustup.rs/ -FileName rustup-init.exe  # Downloads Rustup-init 
-    - rustup-init -yv --default-toolchain %channel% --default-host %host%     # Installs Rust
+    - rustup-init -y --default-toolchain %channel% --default-host %host%      # Installs Rust
     - set PATH=%PATH%;%USERPROFILE%\.cargo\bin;%APPVEYOR_BUILD_FOLDER%        # Adds Rust tools (Cargo, Rustup, etc.) to path
-    - rustc -vV                       # Prints Rust version
-    - cargo -vV                       # Prints Cargo version
+    - rustc -V                        # Prints Rust version
+    - cargo -V                        # Prints Cargo version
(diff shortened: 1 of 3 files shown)


EwoutH:
Does this project still need Rust Nightly btw, or can we switch back to the Stable channel?

MichaelKohler:
Thanks, CI is always good to have. Can you post the Travis and AppVeyor overview pages here?

Also, does AppVeyor run on every commit as Travis does? If so, I'm not sure if we need artifacts for all runs.

Will review the rest tomorrow, looks good to me so far though!

MichaelKohler:
Regarding Rust Nightly, feel free to try, but I think our pinned punkt dependency needs it.

EwoutH:
https://travis-ci.com/EwoutH/common-voice-wiki-scraper and https://ci.appveyor.com/project/EwoutH/common-voice-wiki-scraper on my side.

appveyor.yml
```diff
@@ -0,0 +1,37 @@
...
+    - rustc -V                        # Prints Rust version
+    - cargo -V                        # Prints Cargo version
+    - rustup target add %target%      # Adds target platform to Rust
+
+build_script:
+    - cargo build --release --target=%target%      # Builds file defined in Cargo.toml (default main.rs)
+
+test_script:
+    - cargo test --target=%target% --verbose       # Runs tests in "src" and "tests" folders
+#    - cargo bench --target=%target% --verbose      # Runs benchmarks in "src/lib.rs"
```
MichaelKohler: Can we remove this for now?

appveyor.yml
```diff
@@ -0,0 +1,37 @@
...
+
+artifacts:
```
MichaelKohler: I think for now we should be fine without artifacts and release deployment. If I understand correctly if you remove these lines we could simply merge it without further changes and just need to configure the projects?

EwoutH:
I commented out the test script and the deploy function. I would like to keep the artifacts though (they host it).

appveyor.yml
MichaelKohler: Oh I meant removing just the bench part which was commented out anyway. The tests on Windows might be handy.

MichaelKohler: Oh I meant to remove the bench one as it was commented out anyway. We can keep the tests as they might be useful to run on Windows as well.

MichaelKohler:
Aaaaah, GitHub didn't post my other comment. I meant removing the bench part as it was commented out anyway. We can keep the tests on AppVeyor as it might be useful to also run on Windows.

PR #61, opened 2019-12-30
Mention to a new command to cvtools

Mte90:


README.md
```diff
@@ -98,12 +98,14 @@ You will have to read the ``word_usage.en.txt`` file to decide where you should
...
 ```
 
-Once you know the frequency limit, you can generate your blacklist by running
+Once you know the frequency limit, you can generate your blacklist by running:
 
 ```bash
 python3 ./word_usage.py -i ../common-voice-wiki-scraper/wiki.en.all.txt --max-frequency 80 --show-words-only >> ../common-voice-wiki-scraper/src/rules/disallowed_words/english.txt
 ```
 
+You can use also `--strip-by-apostrophe` that is handy with languages that use `'` in their sentence and reconize more words (european languages use a lot this language feature).
```
MichaelKohler: I'm not sure if the mention of European languages using that is providing much value. I'd leave that off. Also, there is a typo in "recognize" (spelled as "reconize").

Mte90: Mentioning european languages was just a hint to want to experiment with that. Asian languages as example don't use that symbols and is not specific to romance languages because also english us it (I don't know german or russian as example).

MichaelKohler: There are quite a few "non European" languages that use an apostrophe. Let's remove it.

Mte90:
done

PR #64, opened 2019-12-30
Add symbol / character whitelist (fixes #50)

### LINKED ISSUE #50: allowed_symbols as a whitelist instead of disallowed_symbols
When I experimented with the rules for Esperanto it soon became clear that a lot of problems could be avoided by simply excluding most of the extended Latin and Cyrillic alphabet and also some common Chinese signs and Arabic letters. I did this by counting the frequency of letters in all extracted sentences and then add everything I did not want to see to the disallowed_symbols. But this slowed the extraction process down a lot.

It would be much easier to have a whitelist of allowed symbols. This could include the alphabet of the language, a few signs like points, question marks, commas,quotes,... but nothing more. This would regularize the structure of the sentences a lot and could be helpful for many languages.

MichaelKohler:


MichaelKohler:
@stefangrotz any thoughts on this? :)

stefangrotz:
Looks great and will imo solve a lot of problems :) 

Does regex in rust suport alphabet ranges like \p{InArabic} ? This could be very handy here.

MichaelKohler:
> Does regex in rust suport unicode blocks like \p{InArabic} ? This could be very handy here, but it is not widely supported. Your regex in german.tombl looks like a unicode range at least, this will also do the job for all alphabets.

It does support them, just in the following format: `\p{Latin}`. However we probably wouldn't want to use those at least for German, as that would for example also accept `é`. However for other scripts that might be very useful indeed. And if the classes are too broad, unicode ranges can still be used, as you said.

MichaelKohler:
However, this could of course also be combined with the `disallowed_symbols` to filter out some characters again. But I guess it's cleaner to have one regex (which in theory could also be negated as it's run per character).

Issue #14, opened 2019-05-02
Dupe detection


dabinat:
The wiki sentence dump had over 30K duplicate sentences so there should be some kind of duplicate detection built into the scraper tool.

jaumeortola:
Duplicates can be removed just after the whole extraction. 
It would be better to keep a record of all previous sentences during the process, so we don't miss the opportunity to extract 3 new sentences from each article (if possible).

nukeador:
The process is pretty heavy right now and it's taking a lot of hours to complete, @MichaelKohler @Gregoor what would be your recommendation here?

We should probably identify the % of dupes we are getting and if it's low enough maybe doing a post-clean-up instead of making the main process heavier?

MichaelKohler:
https://github.com/Common-Voice/common-voice-wiki-scraper/blob/5fac23bd99db6edf64818af5e1dca5d9b134a4ce/src/extractor.rs#L57

I think there are several issues with that:
* @fiji-flo shouldn't `sentences` be sorted first, as dedup only removes consecutive dupes as per https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup? .. or we could use a HashMap.
* This only dedupes within one "text", the Wiki export has many texts
* I agree that this should happen while choosing, as we otherwise might lose some possible additional extractions.

If we fix the problems above, there will be a perf impact, I can't say how much though off the top of my head.

Gregoor:
Yeah I agree that we should dedupe while picking sentences to maximize the amount of total extracted sentences. And a HashSet is also the cheapest option that comes to my mind.
If we parallelize the process we should be able to more than make up for the cost we have to pay for this. Last time I tried my hand at it, I unfortunately didn't succeed, but happy to restart those efforts.

iLeonidze:
Having issue with duplicates in output file:
```
Любой пользователь может создать новый объект.
Каждый объект может иметь описания на каждом из этих языков.
Также существует блог Викимапии.
Находится в созвездии Кит.
Это самое ближнее к нам скопление галактик.
Находится в созвездии Кит.
Находится в созвездии Кит.
Находится в созвездии Кит.
Это один из наиболее тусклых объектов в списке Мессье.
Два года служил сельским учителем.
Одно из самых известных звёздных скоплений.
Одно из самых известных звёздных скоплений.
Член Союза писателей России.
На небосводе галактика близка к знаменитому густонаселенному скоплению галактик в Печи.
[...]

PR #66, opened 2020-01-05
Initial addition of the Russian language

iLeonidze:
### How many sentences did you get at the end?
`2,118,974` lines on output, `1,871,107` after dedup (`cat wiki.ru.txt | sort | uniq | wc -l`).

### How did you create the blacklist file?
1. A few bad words, which was found in results
2. A lot of abbreviations

### Review
For review please use [sample file](https://drive.google.com/file/d/1H9ZXRO8ENAxPKcaTjcl5GMEHudUj2MBN/view?usp=sharing).

Reviewers:
@swenkal @spartedo @IvanParadox


src/rules/russian.toml
```diff
@@ -0,0 +1,118 @@
...
+  # There is a lot of broken sentences like:
+  # У меня щемило сердце: «Что мы сделаем?
+  # Такой он был веселый, бодрый, нас воодушевлял: „Вы знаете, не бойтесь!
+  # История и культура»".
+  # Пушкин, прочитав поэму, сказал «— Далеко мальчик пойдет».
+  # После увольнения он пытается создать новую жизнь...Жизнь, в свою очередь, строит ему множество «баррикад"
+  # Пушкин, Набоков… мало кого еще можно поставить в этот ряд?
+  # Let's remove them all, we have so many good sentences without problems
+  '«', '»', '"', '„', '“', '”', '‘', '…',
+  '{', '}', '（', '）',
```
MichaelKohler: Even if we have enough sentences when filtering out all of them, there is the option `even_symbols` (https://github.com/Common-Voice/common-voice-wiki-scraper/blob/master/README.md#using-language-rules). I've just added a simplification to the master branch, so you might want to update your branch. Then you can add all these into the `even_symbols` array and the script will make sure to only take sentences with even count of these symbols.

src/rules/russian.toml
```diff
@@ -0,0 +1,118 @@
...
+
+  # 13. ул.Тюменская
+  "[А-Яа-я]\\.[А-Яа-я]",
+
+  # 14. Ставропольский кр.,
+  "\\.,",
+
+  # It is not possible to disallow ALL bad symbols, let's use regex for this
+  # 15. Remove any other bad symbols
+  "[^\\dА-Яа-яёЁ\\s:,.-‑?;!—­‐–'·=’―−”‘]"
```
MichaelKohler: I'd say this disallows a lot of the symbols of `disallowed_symbols`, can you optimize so we only run as many checks as needed? From a very quick glance I'd say we could have `disallowed_symbols` as empty vector, and use this regex for most disallowed symbols. And then together with `even_symbols` it should be fine. I might have missed something though. (see also my other comment)

swenkal:
I checked two segments of "sample file":
1) 169087 - 169300 lines
2) 548455 - 548704 lines

**1 segment:**
169170 line. "Ей, угодниче Божий!"
Weird sentence. It's look like old slavonic lang

169258 line. "В году в карьере актрисы наметился спад."
Missed year numbers. Therefore the meaning of the sentence is lost

169286 and 169288 lines are the same.
"Исполняли функции на общественных началах."

**2 segment** - normal. Did not find some abbreviations, syntax or gramma errors.

iLeonidze:
@swenkal thank you.
> Weird sentence. It's look like old slavonic lang

It's ok, The language is large, it has many different styles of speech, especially since we do not need any particular one.

> Missed year numbers

I found the original Wikipedia article, and actually there were numbers in the original sentence. I don’t know where they went, it’s not my settings, it’s nothing to fix here either.

> lines are the same

This issue already exists #14

iLeonidze:
@MichaelKohler I updated the rules file according to your comments, please check.

nukeador:
Thanks for leading this, super excited to see Russian coming!

For the QA of the outputs, my recommendation would be to create 4 files with 100 randoms sentences each, and get a native speaker (ideally linguist) to review them and comment back with the number of sentences (you can use [this template spreadsheet](https://docs.google.com/spreadsheets/d/1dJpysfcwmUwR4oJuw5ttGcUFYLeTbmn50Fpufz9qx-8/edit#gid=0) for easy evaluation).

On your comment about the blacklist file, it seems you are talking about the rules file, did you tried the process of removing less used words? For example in Spanish we added to the blacklist words with 80 repetitions or less, and that allowed to clean up most complex or foreign words.

Thanks for your efforts!

iLeonidze:
@nukeador first QA already checked ~400 sentences, please see his reply from above. As for others, we are waiting for them to check. Is @swenkal's reply format is ok for you? I suppose spreadsheet is a bit complex for this.

As for blacklist file - I do not see the reason for removing less used words (as I understand you correctly). First of all it is difficult to find out them in 2 million sentences. Second, in Wikipedia articles there is a good quality speech-style, may be there will be a small number of strange sentences, but among two million lines, this is a drop in the bucket.

stefangrotz:
> As for blacklist file - I do not see the reason for removing less used words (as I understand you correctly). First of all it is difficult to find out them in 2 million sentences. Second, in Wikipedia articles there is a good quality speech-style, may be there will be a small number of strange sentences, but among two million lines, this is a drop in the bucket.

Hey iLeonidze, I really recommend using a blacklist, it deletes a lot of foreign words and typos. It is not hard to generate this blacklist, you can find a howto in the readme.md of this reop. If you don't want to loose too many sentences you can decrease the number of repetitions, e.g. you can say you only delete words that are used less than 40 times.

iLeonidze:
@stefangrotz thank you, I did not see at first that there is a script that removes rarely used words.
Of course, I will try and make changes with blacklist as you need it, but can you explain or give me a link, why do I need to get rid of such words?

stefangrotz:
@iLeonidze I can only give you examples from other languages. The list will mainly contain typos. For example "Wikipediaa" instead of "Wikipedia". This will maybe happen 60 times in the complete corpus and this blacklist with every word that appears less than 80 time helps to avoid words like this.

src/rules/russian.toml
```diff
@@ -0,0 +1,78 @@
...
+
+# Apparently, in some places the sentences are cut incorrectly,
+# which is why we get some part of the sentence, and not its entirety.
+# This is required to fix sentences like:
+# с Иваном Галамяном.
+needs_uppercase_start = true
+
+allowed_symbols_regex = "[А-Яа-яёЁ\\s:,.\\-‑?;!—­‐–'·=’―−”‘]"
+
+disallowed_symbols = [
```
MichaelKohler: Note that this won't be used as `allowed_symbols_regex` is set. However, given that those invisible chars should not be allowed by the above regex, I guess that's fine and it can be completely removed?

src/rules/russian.toml
```diff
@@ -0,0 +1,78 @@
...
+needs_punctuation_end = false
+needs_letter_start = true
+
+# Apparently, in some places the sentences are cut incorrectly,
+# which is why we get some part of the sentence, and not its entirety.
+# This is required to fix sentences like:
+# с Иваном Галамяном.
+needs_uppercase_start = true
+
+allowed_symbols_regex = "[А-Яа-яёЁ\\s:,.\\-‑?;!—­‐–'·=’―−”‘]"
```
MichaelKohler: Note that your previous version excluded all quotes and some other characters as it lead to uneven symbol occurances in the sentences. This now allows quotes, so you might end up with sentences with wrong symbols. I'd suggest to add these to the `even_symbols` config option to make sure quotes (and possibly other symbols) are always even.

src/rules/russian.toml
iLeonidze: No. As I found out invisible chars are not detected by regex for some reason. Perhaps because they are part of `\s`. Therefore, they had to be left.

src/rules/russian.toml
iLeonidze: The may difference with the previous version, is that I allowed `‘` symbol - that is ok, in Russian there can be words like "Côte d'Ivoire".

Please, can you provide more info about `even_symbols`, I can't find detailed description about it.

src/rules/russian.toml
MichaelKohler: yeah, that's probably what is happening here. Could you just use the specific character you want for whitespace instead of `\s` or is my Regex knowledge completely failing me here?

src/rules/russian.toml
MichaelKohler: With this regex here, you could potentially get the following sentence: `This is ”uneven and suddenly stops the quote without terminating symbol`

Regarding `even_symbols`, can you tell me what exactly is not clear from the README? Then we can adjust the README to provide better info

MichaelKohler: Did you have a chance to look at this? Is there anything I can help with?

MichaelKohler:
@iLeonidze did you see my last comment? Is there anything I could help you with?

Can I also ask you to do the following, to make sure you can profit from the [automatic sample extraction](https://discourse.mozilla.org/t/scraper-automatic-sample-sentences-extracted-in-pull-request/55217) we just introduced?

* Rename `src/rules/russian.toml` to `src/rules/ru.toml`
* Rename `src/rules/disallowed_words/russian.toml` to `src/rules/disallowed_words/ru.toml`

```
cargo run -- extract -l ru -d path/to/files
```

iLeonidze:
@MichaelKohler Hi, I’m already a little busy for a couple of months, I really want to get back to your help as soon as possible. Thank you for the information, I will try

Nihisil:
@iLeonidze thanks for your work. Is there any update on your side?

PR #94, opened 2020-02-27
Implemented matching opposite characters

SoftBlankie:
Implemented matching by comparing counts of each match


MichaelKohler:
For reference: https://github.com/Common-Voice/common-voice-wiki-scraper/issues/81

Thanks, I'll have a look over the next few days.

README.md
```diff
@@ -85,6 +86,26 @@ The following rules can be configured per language. Add a `<language>.toml` file
...
+  ["(", "]"],
+  ["[", ")"],
+]
+```
+
+This matches all occurrence of `„` with `“`, all occurrence of `(` with `]`, all occurrence of `[` with `)`.
+
+```
+Input: This is „a test“ and (another one]
+Output: Valid
```
MichaelKohler: I see how this could be useful to show that it doesn't need to be semantically the opposite character. However, on the first look this might be confusing. Maybe we could use ( and ), [ and ] for the example and add a second one which explains that what you did here is in theory possible as well?

PR #99, opened 2020-03-04
DE rules: add 'Klasse' to filter

MichaelKohler:


MichaelKohler:
@jf99 I just had to create a PR to document something. If you think this would be useful, I'd be happy to also merge it instead of just closing it.

benekuehn:
I think it would be helpful!

PR #101, opened 2020-03-12
Enable Full Extraction through GitHub Actions

MichaelKohler:
This enables full extraction through GitHub Actions. If we pass `--full-extraction=<locale>` in the merge commit after accepting a rule, it will automatic generate the full extraction file. Then we just need to download that sample and add it to the voice-web repo.

Example for ca: https://github.com/MichaelKohler/common-voice-wiki-scraper/runs/501975370?check_suite_focus=true (one big dump file, but Actions can still handle it)
Example for es: https://github.com/MichaelKohler/common-voice-wiki-scraper/runs/503056319?check_suite_focus=true (many smaller dumps we iterate over)

Caveat: if there are multiple dump files, we run the extractor multiple times and therefore might end up with duplicates. I'll add a deduplicate step in a later PR.

@nukeador can you have a look at the Spanish run and check if there is something obviously wrong so we have double checked this PR?

### DIFF
--- a/.github/workflows/extraction.yml
+++ b/.github/workflows/extraction.yml
@@ -1,6 +1,6 @@
-name: Sample Extraction
+name: Extraction
 
-on: [pull_request]
+on: [pull_request,push]
 
 jobs:
   extract:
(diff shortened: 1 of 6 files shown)


MichaelKohler:
> Caveat: if there are multiple dump files, we run the extractor multiple times and therefore might end up with duplicates. I'll add a deduplicate step in a later PR.

I've added a fix for this now here as well.

PR #105, opened 2020-06-01
add translate option

yanganto:
- with `--trans` or `-t` the traditional Chinese will translate to simplify Chinese
- default will not translate

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -30,6 +30,12 @@ where
                         .takes_value(true)
                         .number_of_values(1)
                         .help("input dir to glob"),
+                )
+                .arg(
+                    Arg::with_name("trans")
+                        .short("t")
+                        .long("trans")
+                        .help("auto translate tradistional chinese to simplify chinese"),
                 ),
         )
         .get_matches_from(itr)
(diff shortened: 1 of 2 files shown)


yanganto:
This feature will merge to [Mozilla Hong Kong](https://github.com/mozillahk/)  as following PR.
https://github.com/mozillahk/cv-sentence-extractor/pull/1

You may close this PR, if you are not want this.  Thanks.

MichaelKohler:
I think it would indeed be great to integrate this into the "mandarin" branch here, of course nicely configurable so it's not HK specific. I'd suggest to work on that in your  repo, and then once that works we can do a PR into this repo here? Does that sound reasonable?

Thanks!

yanganto:
Got it.  Sorry for opening a PR in rude.  This PR is implemented by the request from the common voice in Hong Kong telegram group.   Once the PR of  Mozilla Hong Kong accepted,  I will open the PR back to here.  Thanks.

MichaelKohler:
No worries! Is that the only change needed? I would gather all changes required together in the HK repo, and then once everything is done and works open a PR here.

yanganto:
There are at least three features need to be implemented.  This is just the first.  I will make PR here when all the features in completed.  Thanks for remind.

PR #124, opened 2020-08-27
Add rules and blockwords for Lithuanian language --full-wiki-extraction=lt

mjurkus:
~130 000

**How did you create the blacklist file?**
As described in readme. Filtering out works frequency 10 and less. After manually inspecting it I noticed really great words, that should not be blocked.
I'm not sure if I should reduce it more?
Going manually over 900 000 entries doesn't seem like a productive way to spend time, especially if it might be regenerated next time by someone else.

**Get at least 3 different native speakers (ideally linguistics) to review a random sample of 100-500 sentences and estimate the average error ratio and comment (or link their comment) in the PR.**

Created 4 random samples of 200 sentences: `shuf -n 200 -o sample.txt wiki.lt.txt`

Made a document with those samples. I'll review one and share others with native speakers (I even might find linguists): [Google Sheet Document](https://docs.google.com/spreadsheets/d/1sduhQ-oUbpeSzJORsgizL3_xBhtpIdcbrhodx4xMDh4/edit?usp=sharing)
I'll also discuss if the blacklist file could be improved.

### DIFF
--- a/src/rules/lt.toml
+++ b/src/rules/lt.toml
@@ -20,9 +20,11 @@ disallowed_symbols = [
 ]
 abbreviation_patterns = [
 	"[A-ZĄČĘĖĮŠŲŪŽ]+\\.",
-	"[A-Z]+\\.*[A-Z]",
+	"[A-ZĄČĘĖĮŠŲŪŽ]+\\.*[A-ZĄČĘĖĮŠŲŪŽ]",
+	# for UAB, LŽS and similar
+	"[A-ZĄČĘĖĮŠŲŪŽ]{2,}",
 	# coommon abbr
-	"kt\\.|pr\\.|[Vv]yr\\.|[Mm]ot\\.|žr\\.|Šv\\.|[Šš]vč\\.|mėn\\.|el\\.|Dž\\.|įv\\.|dab\\.|gel\\.|ež\\.|[Nn]r\\.",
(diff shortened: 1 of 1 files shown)


MichaelKohler (commented):
Code generally looks good to me so far. For the blocklist: the limit really depends on how readable the resulting sentences are. For other languages pushing up that limit has resulted in a lot of hard to speak aloud sentences being thrown out which helps users while recording sentences. Let's see in your review how many hard sentences there are.

mjurkus:
@MichaelKohler 2 native speakers (one of them was me) and 1 linguist reviewed given sentences. Do you need a comment from them?
Error rate is 5%, 2%, 1%, 1% for each set respectively.

Main issues are English names, location, brand other names not adapted to Lithuanian language.
Also Romain numeral "I" that wasn't caught by regexp.

MichaelKohler (approved):
> Do you need a comment from them?

Nah, I think that's fine.

> Main issues are English names, location, brand other names not adapted to Lithuanian language.

Increasing the blocklist threshold would probably help here, but these error rates look good enough to me.

>  Also Romain numeral "I" that wasn't caught by regexp.

That heavily depends on the language whether it's easy to capture that or not. If you want to try to find a regex that works for Lithuanian, go for it. But I won't make you do it.

So yes, generally I'd say that error rate looks good. Tell me if you want to do changes to the blocklist and regex for the remaining issues. If not, I'll be happy to start the official export.

Thank you!

mjurkus:
@MichaelKohler updated rules. I think we can merge now.

MichaelKohler:
Thank you!

MichaelKohler:
The extraction on merge failed, I fixed the bug and retriggered the full extraction here: https://github.com/Common-Voice/cv-sentence-extractor/runs/1046216702?check_suite_focus=true

I will create the PR with the sentences in the common-voice repo and CC you as well.

PR #131, opened 2020-11-01
Fix regex rules in hu.toml

xTibor:
"ű" and "Ű" are also valid Hungarian letters.

Source: Rules of Hungarian Orthography, 12th edition (AkH.12), section 4. and section 10.


MichaelKohler (approved):
Thanks!

PR #134, opened 2021-02-16
Initial Galician rules and blocklist

xosecalvo:
== How many sentences did you get at the end? ==
Extracted sentences: 1 221 442
Number of sentences at the end (in wiki.gl.txt): 41 407

== How did you create the blacklist file? ==
The list with word frequencies had a threshold of 50. I preferred to do more work on the filtering rather than miss useful words. Galician place names were removed using a list. No bad words were included, as none were found on inspection, but quite a few words were added as they were found during analysis, mostly foreign names. This has taken a long time.

== Review ==
Three native speakers, all of them linguists, have reviewed a sample of the sentences. I am one of them. We agreed on looking out for these problems:
1) Misspellings
2) Typos
3) Grammar mistakes
4) Foreign, unusual proper names
5) Meaningless, lacking co-text 

We ran several iterations. We didn't catch many mistakes for reasons 1 and 2, and, at first, most were of type 4. These were caught with the blocklist and by barring some letters and letter combinations. Eventually, we arrived at these results:

1) Me:
Examined sentences from 10 000 to 10 501
Of these 500 sentences:
· 3 should be removed because they were missed by the filter (2 foreign words (2) + 1 abbreviation before a stop)
· 3 should be removed because they are originally wrong in the Galician Wikipedia
Therefore, the error rate I found was 1,2%.

2) M. B.
Looked at 1,200 sentences (400 to 1000 and 14 000 to 14 601). She rejected 17, 7 of them due to their including foreign words, that is, an error rate of 1,42%.

3) A. M.
Automatic analysis with hunspell: 0,63%
500 including other problems: (typos, grammar mistakes, meaning): 10 (2% error rate)

We deem these results to be acceptable.


src/rules/gl.toml
```diff
@@ -0,0 +1,40 @@
+min_trimmed_length = 3
+min_word_count = 2
+max_word_count = 14
+min_characters = 1
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = true
+disallowed_symbols = [
```
MichaelKohler, line 10: You might want to also have a look at `allowed_symbols_regex`, might make this easier.

MichaelKohler:
Thanks for working on this! The rules generally look good to me! I have a few questions though:

```
Extracted sentences: 1 221 442
Number of sentences at the end (in wiki.gl.txt): 41 407
```

This seems fairly low to me. Do you happen to know what the major contributors of "wrong" sentences are? Is that all due to words in the block list? Note that we can't go back and re-do the export with different rules once it's done, in the future we can only scrape pages that were created after we ran the extractor. Getting as many sentences as possible on this initial run, while of course still maintaining quality, is crucial here.

> The list with word frequencies had a threshold of 50. I preferred to do more work on the filtering rather than miss useful words. Galician place names were removed using a list. No bad words were included, as none were found on inspection, but quite a few words were added as they were found during analysis, mostly foreign names. This has taken a long time.

Did you experiment with lower threshold values? I can't really recommend a number here though, other languages have used anything between 4 and 100.

> We deem these results to be acceptable.

That's way below other language's initial review. This might indicate that you might be able to lower the threshold to get more words.

xosecalvo:
Thanks for your suggesting allowed_symbols_regex. It looks promising as it reduces the need for disallowing so many symbols. I'll be looking into it.

As for the low number of pages - yes, it's pretty low, we're aware of that. We weren't sure what an 'acceptable' error rate would be, and we've striving to make it as low as possible. Also, since we can't reject wrong results once they're in, we don't want to waste people's time making them read meaningless sentences or having lots of sentences in the "clip cemetery" because of typos.

Most sentences were rejected because they included foreign names. With a threshold of 50, the first 10 words are these:
abandonouno
abanico
abbásida
acelerouse
acenos
acontece
adaptadoras
adecuado
*adele
administrando
There's just one with a problem. With a threshold of 40 we got this:
*abarth
aberracións
*acanthamoeba
acepte
*achievement
*ackroyd
actuará
acuosos
*adar
*adenauer
Not that they're wrong, but one of them is English, 4 are foreign names and the other one (acanthamoeba) is not in the dictionary (yet). Do you think we're being too strict?

MichaelKohler (approved):
I see, looks like 50 is a good treshold value. I'm approving this PR as all the requirements are fulfilled. However, I wouldn't say no to you trying to get more sentences without majorly decreasing the quality. An error rate below 5% is generally acceptable.

Maybe @ftyers has a good idea.

ftyers:
I would say it's quite worthwhile having foreign names and placenames in the corpus. People may have different ways of pronouncing them, but if the plan is to make a general ASR system, then proper names are likely to be useful. I know in the Catalan one they include a lot of placenames in Catalan as they can sometimes have unpredictable pronunciation. Other than that this work looks good.

xosecalvo:
Thanks to @MichaelKohler and @ftyers for your suggestions. I've changed gl.toml completely, this time using allowed_symbols_regex. This makes things so much easier! The number of sentences has jumped up to **76 603** but the best side effect is that the blocklist, which was 6,7MB, is now only 201,8kB!! We're happier with the results, particularly because we can make use of the blocklist for, for instance, improving the Galician language dictionary.

MichaelKohler (approved):
Changes look good to me! I'll check tomorrow why the example extraction failed, it's not an issue with your rules file though.

xosecalvo:
I've been sent the new release of the Galician dictionary. I'll run this again.

MichaelKohler:
The Github Action error has now been fixed in the main branch. Would be great if you could merge the latest main branch into your branch and push it.

xosecalvo:
I've generated the blocklist by running the extractor multiple times, sending the errors to the blocklist using the new Galician dictionary and tweaking gl.toml even further. This has allowed us to reach **91 821** sentences, while keeping gl.txt at just 268,7kB. The overall quality of the texts has improved, too.

xosecalvo:
Is there anything keeping this merge back? Do I need to do anything else or is there anything I can do to have this moved forward?

MichaelKohler (changes requested):
Oh, sorry for not getting back to you on this. Could you remove the `gl.toml` file in the root folder? The one in `src/rules` is enough. Then we can merge this. Once merged the Github Action will run the export and we'll create the PR with the new sentences in the Common Voice repository.

Thank you!

xosecalvo:
Oh, my! I hadn't noticed I had sent that file there, I'm sorry.

MichaelKohler (approved):
Thank you!

MichaelKohler:
PR with the automated extract created here: https://github.com/common-voice/common-voice/pull/3053

Issue #138, opened 2021-04-17
Documenting the order of execution for language rules


bact:
Currently, it is not obvious on the order of rules. I cannot find any mention in the document.

It will be very useful for people who develop the rules to know the order.

For example, to know if `replacements` will be executed before `disallowed_words` or `broken_whitespace` or not.

MichaelKohler:
Apart from `replacements`, what other rules would be useful to know the order of? I'd argue that apart from `replacements` it shouldn't matter.

Definitely agree that `replacements` should mention that this runs before any other rules.

bact:
If it's the case that `replacements` run before any other rules, I think you are correct - the order probably doesn't matter for the rest. 

I was worried about length-related rules (like `min_trimmed_length`, `min_word_count`), but it's clear now if `replacements` runs the first.

MichaelKohler:
Definitely a good point, added!

bact:
Yeah, I feels like `replacements` is actually different from other rules. It is an action. While others are constraints.

MichaelKohler:
I definitely shouldn't mark issues as "good first issue" and then do them myself, sorry! :)

>  feels like replacements is actually different from other rules. It is an action. While others are constraints.

Can't disagree with you there, but moving it out of the rules files makes it way more complicated than it needs to be :)

PR #140, opened 2021-04-17
Note on order of rules: replacements runs first

bact:
This will fix part of #138


MichaelKohler:
Oops, just did that in https://github.com/common-voice/cv-sentence-extractor/commit/d24d060e443629ab48342051e960b12a071234fa. Thanks though!

PR #141, opened 2021-05-10
Changes for italian

Mte90:
Looking to the reported sentences (https://gist.github.com/Mte90/e3e67b067da77cfa8d41ae6d1b1b9c87) we have 2 big issues:

* People name in a sentence that are not italian that is just the sentence, so forcing a final punctuation end should avoid that case
* External alphabets that creates issues

Another idea is to put as minimum for sentences 2 words but I don't think that will works with names of three words.


Mte90:
What are the next steps for the re-run?

MichaelKohler:
This has been answered here: https://matrix.to/#/!sRgqihuSaMJkWPZPSN:mozilla.org/$K7S4fnluSioqYg3WazehE6y67xTT32mHByJ7sgQkHlA?via=mozilla.org&via=matrix.org&via=t2bot.io

If there are any unanswered specific questions feel free to ask them specifically. For this PR here, no matter of the timeline if a re-run, you will need to go through validation again to make sure these changes don't break anything. Let's use the same process as described in the README (even small changes can have a big effect on this, keep also an eye out on a possible decrease in number of sentences due to possible over-filtering).

Mte90:
Took 3 hours of processing and generated 561074 sentences and at first look they are very better compared with the oldest.

[wiki.it.zip](https://github.com/common-voice/cv-sentence-extractor/files/6510158/wiki.it.zip)

For the validation is required to have some volunteers writing here that they are good?

MichaelKohler:
Do you know if the initial export included the blocklist as well? It was done a week after the blocklist got merged, so I'd say yes. The commit message to CV also suggests so as it mentions < 80 repetitions.

If that's the case, the export now has > 300k less sentences, even though the Wikipedia grew over that time. There was also https://github.com/common-voice/cv-sentence-extractor/pull/55 merged before this here which might have had an impact as well. In the end I'm a bit concerned about a "loss" of 33% of sentences, even if it improves the quality overall. Are the current sentences in CV really that bad?

> For the validation is required to have some volunteers writing here that they are good?

Yes, please. Let's treat this like a new rules file (check the README, there is also a template you can use) as minor changes can have a big impact.

Mte90:
Yes the first export had it because we were trying to clean it most as possible.

About the quality yes, we had a lot of troubles and now we are forcing only for the alphabet and sentences that ends with a dot as we had a lot of reports of wrong sentences and we avoid al those names.
I was concerned too that are less sentences compared to before.
I saw also that now doesn't include words from english as there aren't `yjkwx` letters in the alphabet. This will help a lot on reviewing.

Maybe run a new rescan and merge them without the duplicates? As they are picked randomly as i can remember can be a solution?
For the other steps I will check them.

MichaelKohler:
> Maybe run a new rescan and merge them without the duplicates? As they are picked randomly as i can remember can be a solution?

Not sure I understand. If you're suggesting to run it on all articles and merge it with the existing sentences, that won't work as we can't have more than 3 sentences per article.

Mte90:
We were ignoring also the grave accents and `yjkwx`.

src/rules/it.toml
```diff
@@ -4,16 +4,11 @@ max_word_count = 16
...
-  'â', 'ă', 'ș', 'ç', 'Å', 'ý', 'ö', 'ü', 'ā', 'ö', '%', '&',
-  'α', 'β', 'Γ', 'γ', 'Δ', 'δ', 'ε', 'ζ', 'η', 'Θ', 'θ', 'ι', 'κ',
-  'Λ', 'λ', 'μ', 'ν', 'Ξ', 'ξ', 'Π', 'π', 'ρ', 'Σ', 'σ', 'ς', 'τ',
-  'υ', 'Φ', 'φ', 'χ', 'Ψ', 'ψ', 'Ω', 'ω', '©'
-]
+allowed_symbols_regex="[A-Za-zìàèòùáéíóúÁÉÍÓÚÀÌÒÈ;‚'–\\. \"\\?!]"
+disallowed_symbols = []
 broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;", " :"]
-disallowed_words = []
+disallowed_words = ['Jr', 'vs', 'vol', 'Av', 'Blvd', 'aprox', 'Cd', 'Cen', 'Ant', 'etc', 'art', 'Ltd', 'Co', 'esq', 'Corp', 'Dra', 'Prof', 'Urb', 'Bs', 'Mr', 'Mrs', 'subsp', 'p.e','s.s', 'Av', 'Blvd', 'Arq', 'St','Sta', 'S.', 'Sr.', 'Dr.' ]
```
Sav22999, line 13: Why "Prof" is in the disallowed?
It's an "abbreviation" (actually it's an apocope -> troncamento)

It's like "signore"->"signor", etc. both forms are correct

The "Sta" abbreviation could mean also the verb "stare" (Io sto, tu stai, egli/ella sta, ...) Present, 3rd singular

Sav22999: I would add to "disallowed words" also "I.", "II.", "III.", ..., "XX." (I think until the 20th is enought).
Maybe also "1°", "2°", ..., "20°"

Sav22999 (changes requested):

Mte90:
Anyway with the latest changes also if not finished found more sentences compared to the first run of this PR.

Mte90:
With the latest changes we are getting 919014 sentences.
We are still testing it and I will share there when it's ready.

src/rules/it.toml
```diff
@@ -10,5 +10,5 @@ needs_uppercase_start = true
 allowed_symbols_regex="[A-Za-zìàèòùáéíóúÁÉÍÓÚÀÌÒÈ;‚'–\\. \"\\?!]"
 disallowed_symbols = []
 broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;", " :"]
-disallowed_words = ['vs', 'vol', 'blvd', 'ltd', 'aprox', 'ecc', 'art', 'Co', 'Urb', 'Mr', 'subsp', 'p.e', 's.s', 'SS', 'H.', 'St', 'S.', 'Sr.', 'Dr', 'I.', 'II.', 'III.', 'IV.', 'V.', 'VI.', 'VII.', 'VIII.', 'IX.', 'X.', 'XI.', 'XII.', 'XIII.', 'XIV.', 'XV.', 'XVI.', 'XVII.', 'XVIII.', 'XIX.', 'XX.' ]
+disallowed_words = ['vs', 'vol', 'blvd', 'ltd', 'aprox', 'ecc', 'art', 'Co', 'Urb', 'Mr', 'Inc', 'subsp', 'p.e', 's.s', 'SS', 'H.', 'St', 'S.', 'Sr.', 'Dr', 'I.', 'II.', 'III.', 'IV.', 'V.', 'VI.', 'VII.', 'VIII.', 'IX.', 'X.', 'XI.', 'XII.', 'XIII.', 'XIV.', 'XV.', 'XVI.', 'XVII.', 'XVIII.', 'XIX.', 'XX.' ]
```
Sav22999, line 13: Add also
"Mr.", "Mr", "Mrs.", "Mrs", "Dr.", "Dr", "Jr.", "Jr"

Sav22999 (changes requested):

src/rules/it.toml
Sav22999: Also "secc", "sec"

MichaelKohler:
> I am wondering if needs_punctuation_end includes just . or also ?!.

If that option is set and the sentence ends with an alphabetic character it's invalid. So yes, any of those are counted as valid. This would also count : as true, that's why there is the second option to disallow that (may_end_with_colon).

Mte90:
839944 sentences: [wiki.it.zip](https://github.com/common-voice/cv-sentence-extractor/files/6522382/wiki.it.zip)

Review file: https://docs.google.com/spreadsheets/d/1RGWRvPmZJNPE92MSPa25UlWhak2J9PEymaA0WIb372c/edit?usp=sharing

Sav22999:
I reviewed the sentences in the spreadsheet (and I've also checked randomly some sentences in the wiki.txt zip).

Sav22999 (approved):

MichaelKohler (approved):

PR #153, opened 2021-07-22
Add Initial rules and blocklist for Swahili

OmondiKevin:

- After a successful run and tweaks from the original outputs, this code generates 9189 sentences of a good accuracy upon review. 
- I have posted sample 1000 sentences generated here: [check my GitHub gist here](https://gist.github.com/OmondiKevin/4958dfdf0287392041a3b47601ce1625) any Swahili speakers out there is welcomed to view this and provide comments.

**How did you create the blocklist file?**

- Based on the guidelines provided in the readme section and filtering out words with a frequency of 80 and less. The frequency of 80 was giving us good sentences that helped us come up with some good replacements as in our sw.toml file. Our min word count was also 3. 

**Results from the reviews**

- We had 3 contributors, they accuracy rates were 99%, 100% and 94% which turned out to be good. The review results and the comments can be viewed here [Check Reviews on the sample sentences here](https://docs.google.com/spreadsheets/d/1QJd4KLcRyhVcagLWQzknJ02-7JZOSU60h0WJ6EkvrUU/edit?usp=sharing)


src/rules/sw.toml
```diff
@@ -0,0 +1,33 @@
...
+  'α', 'β', 'Γ', 'γ', 'Δ', 'δ', 'ε', 'ζ', 'η', 'Θ', 'θ', 'ι', 'κ',
+  'Λ', 'λ', 'μ', 'ν', 'Ξ', 'ξ', 'Π', 'π', 'ρ', 'Σ', 'σ', 'ς', 'τ',
+  'υ', 'Φ', 'φ', 'χ', 'Ψ', 'ψ', 'Ω', 'ω',
+]
+broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;", " :", "„ ", " “", "„“", "“-", "-,", ".“", " -", "- ", "?-"]
+abbreviation_patterns = ["[A-Z]+\\.*[A-Z]"]
+matching_symbols = [
+  ["„", "“"],
+  ["(", ")"],
+  ["[", "]"]
```
MichaelKohler, line 22: Are these actively used in the text? Is it clear to readers what these mean so they don't stumble over it?

src/rules/sw.toml
```diff
@@ -0,0 +1,33 @@
+min_trimmed_length = 3
+min_word_count = 3
+max_word_count = 14
+min_characters = 0
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = false 
+disallowed_symbols = [
```
MichaelKohler, line 10: Might be worth to look into `allowed_symbols_regex` to turn this around. https://github.com/common-voice/cv-sentence-extractor#using-language-rules

MichaelKohler (commented):
Thanks!

According to https://en.m.wikipedia.org/wiki/List_of_Wikipedias there are around 65k articles. Therefore I'm a bit surprised that this resulted in less than 10k sentences. If you run the extractor with `--no_check`, how many sentences do you get?

src/rules/sw.toml
sdenton4: Hi, Michael;
We ended up erring on the conservative side; there's a fair number of English words and place-names with pretty high frequency. Clicking through the 'random pages' feature also indicates that a decent proportion of the articles are stubs about places; a few sentences with low-frequency place-names, which will tend to be filtered.

https://sw.wikipedia.org/wiki/Maalum:UkurasawaBahati

src/rules/sw.toml
sdenton4: Hi, Michael; The matching rules for parentheses and such are the same for written Swahili and English, so pulling these forward from the base rules should be fine.

MichaelKohler: Fair!

src/rules/sw.toml
MichaelKohler: Thanks for the answer. So you feel confident that 9189 is the maximum while still keeping quality to a good level? If so, I'm fine with that, but I want to make sure we make the best of it, as we can't re-run the extract on these articles more than once.

sdenton4: Yeah, we ran the extractor locally at (iirc) frequencies 60, 70, 80, 90, and 100, and then examined the sentences in the diffs.

Suppose we have D sentences in the diff between frequencies F and F', and N_F total extracted sentences with frequency F.  We examine the sentences in the diff, and estimate that they're bad with probability p. Then p * D / N_F is a lower bound on the ratio of bad sentences in the full set of sentences at frequency F, so we want p * D / N_F < 0.05. We settled on frequency 80 as the sweet spot for getting past lots of English words and place names which were dragging down quality. I think Kevin has the spreadsheets where we did the analysis; my memory is that we pick up a couple thousand sentences if we dial back to frequency 70, but would need to develop further word block-lists to bring the miss-rate under control.

MichaelKohler: Wow, nice data based approach!

MichaelKohler (approved):
I'll merge this in a bit and let the extract run. I'll create a PR in the common-voice repository with the resulting sentences. Thanks!

PR #157, opened 2021-10-12
Improve DE rules file

stefangrotz:
I thought there were more things to improve, but this is the only thing I found. After that imo we are ready for a second extraction.


MichaelKohler:
Every commit on a PR automatically triggers a sample extraction on GitHub (second one in the row below)

image

If you click on "Details", and then "Artifacts", you can download it:

image

In this case here it extracted 15k sentences, which is pretty nice to do validation on without having to run anything locally.

src/rules/de.toml
```diff
@@ -7,12 +7,13 @@ quote_start_with_letter = true
...
 needs_letter_start = true
 needs_uppercase_start = true
-allowed_symbols_regex = "[\u0020-\u007BäöüßÄÖÜ„“‚‘’–\\.]"
+allowed_symbols_regex = "[\u0020-\u007BäöüßÄÖÜ„“‚‘’–\\.?!]"
 disallowed_symbols = []
 broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;"]
 
 matching_symbols = [
-  ["„", "“"]
+  ["„"«, "“"»],
```
MichaelKohler:
```suggestion
  ["„", "“"],
```
This fixes the syntax error. And the other type of quotes is generally not allowed anyway.

MichaelKohler:
I'd say this generally looks good to me. Couldn't mark anything as I don't have edit permissions, but that's fine.

Two thoughts:

* Might be worth adding `"` to `even_symbols` as it appears in a few sentences without a counter part. However I will check why that even shows up, as in theory it's not an allowed symbol (though it probably should be).
* Might be worth adding `ggf\\.$` to the other pattern as it should not be at the end of a sentence

src/rules/de.toml
```diff
@@ -7,12 +7,13 @@ quote_start_with_letter = true
 needs_punctuation_end = false
 needs_letter_start = true
 needs_uppercase_start = true
-allowed_symbols_regex = "[\u0020-\u007BäöüßÄÖÜ„“‚‘’–\\.]"
+allowed_symbols_regex = "[\u0020-\u007BäöüßÄÖÜ„“‚‘’–\\.?!]"
```
MichaelKohler, line 10: Ah, I see why " is allowed, it's part of this unicode range. Then we probably indeed should add it to the even count rule. Additionally ! and ? are already part of this range, so it doesn't need to be added. However that range also includes some characters that we probably not want, such as $, % and others. https://unicode-table.com/en/#0021

Sorry, should have had a closer look before...

src/rules/de.toml
```diff
@@ -7,12 +7,13 @@ quote_start_with_letter = true
...
 disallowed_symbols = []
 broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;"]
 
 matching_symbols = [
-  ["„", "“"]
+  ["„", "“"],
+  ["(", ")"]
 ]
```
MichaelKohler, line 18: Let's add `even_symbols = ['"']` here.

src/rules/de.toml
MichaelKohler: Basically. What I would suggest here is the following:

```
allowed_symbols_regex = "[\u0020A-za-zäöüßÄÖÜ\"„“‚‘’–\\.?!()]"
```

stefangrotz:
I corrected the allowed symbols, added even symbols and added abbreviation ggf. But now a syntax error in broken_whitespace is highlighted. any idea whats wrong there?

src/rules/de.toml
```diff
@@ -7,20 +7,22 @@ quote_start_with_letter = true
 needs_punctuation_end = false
 needs_letter_start = true
 needs_uppercase_start = true
-allowed_symbols_regex = "[\u0020-\u007BäöüßÄÖÜ„“‚‘’–\\.]"
+allowed_symbols_regex = "[\u0020A-za-zäöüßÄÖÜ\"„“‚‘’–\\.?!()]"
```
MichaelKohler, line 10:
```suggestion
allowed_symbols_regex = "[\u0020A-Za-zäöüßÄÖÜ\"„“‚‘’–\\.?!()]"
```

I'm sorry, I made a mistake when I suggested this. :( Today is not my day somehow..

stefangrotz:
Thanks. I updated the evaluation sheet with a new sample: https://docs.google.com/spreadsheets/d/1WRjnz-rWeJwZenl5V71At5Q8CQdBk4A5cIKpuSXf8Ds/edit#gid=0

guerda:
I would also add "Dr." to the abbreviation list, as discussed in [discourse](https://discourse.mozilla.org/t/tokenizer-trennt-haufig-satze-auf/87288?u=guerda).
See my PR here: https://github.com/stefangrotz/cv-sentence-extractor/pull/1

stefangrotz:
Hey @guerda thanks, I added Dr. and Prof. , I am sure there are a lot of other possible abbreviations that we could add, I will think about this and collect more common abbreviations.

@MichaelKohler Re-running the blocklist is definitely a good idea after so much time. Maybe it would be good to put the numbers a little higher to avoid uncommon words.

guerda:
> Hey @guerda thanks, I added Dr. and Prof. , I am sure there are a lot of other possible abbreviations that we could add, I will think about this and collect more common abbreviations.

Great, thank you! I'm already maintaining a small list of abbreviations which I can imagine and one with words which appear after a number and occur in the sentences I speak or listen to.

MichaelKohler:
I reviewed a few more sentences in https://docs.google.com/spreadsheets/d/1Q_IO0GLg8vdjyYEMlCctL5sB8QYMEPBhbYgUF6aheyM/edit#gid=0 . What I noticed is that we probably could add "Abs." to the replacements as well.

MichaelKohler:
I've merged this as the extraction pipeline by now seems to have some issues. I will create a follow-up PR for needed changes and an additional review.

PR #160, opened 2021-10-14
Add Initial rules and blocklist for Igbo (ig)

chrisemezue:

After a successful run and tweaks from the original outputs, as well as help from @MichaelKohler and @stefangrotz we got  this code generates **2165** sentences of a good accuracy upon review. 

[Here](https://github.com/chrisemezue/Igbo-Speech-Dataset) are all the steps I took.

**How did you create the blocklist file?**

Based on the guidelines provided in the readme section and filtering out words with a frequency of 20 and less. The frequency of 80 reduced our sentence to ~400. We checked and found that 20 was also giving us good sentences. I did not tweak the **ig.toml** file much.

**Results from the reviews**
Spreadsheet of 800 random sentences has been reviewed. Results of review are [here](https://docs.google.com/spreadsheets/d/1sjtQh2-0FG2juN3e9BnM3wRSvoICcXvqx7z7L_LuZos/edit?usp=sharing).


src/rules/ig.toml
```diff
@@ -0,0 +1,23 @@
+min_trimmed_length = 3
+min_word_count = 3
+max_word_count = 14
+min_characters = 0
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = false 
+disallowed_symbols = [
```
MichaelKohler, line 10: `disallowed_symbols` got replaced by `allowed_symbols` as it better filters out non-useful characters. We are currently in the progress of changing that in the English file as well, you can take a sneak peek here: https://github.com/common-voice/cv-sentence-extractor/pull/159/files#diff-0ba36324ff3a950efd76d64d802802688fdfe6bea026934e7f3cb9cd3140c305R12

src/rules/ig.toml
```diff
@@ -0,0 +1,23 @@
...
+  '<', '>', '+', '*', '\', '#', '@', '^', '[', ']', '(', ')', '/',
+  'é', 'è', 'à', 'ç', 'Å',
+  'α', 'β', 'Γ', 'γ', 'Δ', 'δ', 'ε', 'ζ', 'η', 'Θ', 'θ', 'ι', 'κ',
+  'Λ', 'λ', 'μ', 'ν', 'Ξ', 'ξ', 'Π', 'π', 'ρ', 'Σ', 'σ', 'ς', 'τ',
+  'υ', 'Φ', 'φ', 'χ', 'Ψ', 'ψ', 'Ω', 'ω',
+]
+broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;", " :", "„ ", " “", "„“", "“-", "-,", ".“", " -", "- ", "?-"]
```
MichaelKohler, line 17: `"„“", "“-", "-,", ".“"`and `"?-"` might better fit into the `other_patterns` rule, but that's purely an improvement of understanding the rules file later on.

src/rules/ig.toml
```diff
@@ -0,0 +1,23 @@
+min_trimmed_length = 3
+min_word_count = 3
+max_word_count = 14
+min_characters = 0
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = false
```
MichaelKohler, line 9: Looking at random Wikipedia articles it seems to me that the correct grammar would be with an uppercase letter at the start of a sentence. I might be wrong, but something you might want to look into :)

src/rules/ig.toml
```diff
@@ -0,0 +1,23 @@
...
+  'α', 'β', 'Γ', 'γ', 'Δ', 'δ', 'ε', 'ζ', 'η', 'Θ', 'θ', 'ι', 'κ',
+  'Λ', 'λ', 'μ', 'ν', 'Ξ', 'ξ', 'Π', 'π', 'ρ', 'Σ', 'σ', 'ς', 'τ',
+  'υ', 'Φ', 'φ', 'χ', 'Ψ', 'ψ', 'Ω', 'ω',
+]
+broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;", " :", "„ ", " “", "„“", "“-", "-,", ".“", " -", "- ", "?-"]
+abbreviation_patterns = ["[A-Z]+\\.*[A-Z]"]
+matching_symbols = [
+  ["„", "“"],
+  ["(", ")"],
+  ["[", "]"]
```
MichaelKohler, line 22: if there are also `"` in the sentences, then it might make sense to add the following as well:

```
even_symbols = ['"']
```

src/rules/ig.toml
```diff
@@ -0,0 +1,23 @@
+min_trimmed_length = 3
+min_word_count = 3
+max_word_count = 14
```
MichaelKohler, line 3: It seems to me that sentences have a rather large number of words. Not knowing the language I have no idea if those are just generally long sentences, but it at least seems that words generally are shorter in Igbo.

How many words on average can you speak out load in normal pace within around 5 seconds? How many in 7 seconds?

Bumping up the number here might make sense depending on that.

src/rules/ig.toml
chrisemezue: You may be right. many of the extracted sentences did start with an uppercase letter

chrisemezue (commented):
I have taken all your pointers into consideration. 
The most important edits I made (and realizations) are:
1. Increasing the `max_word_count` to 15 which is readable in ~5-8seconds
2. Greatly trimming down the disallowed words in `ig.txt`. On careful analysis, I realized that due to the code-switching, dialectal and diacritical complexities of Igbo, many of the infrequent words were actually very correct (contrary to the rule that less frequent words are not good). Plausible explanations for their infrequency are due to:
i. words in English having a number of widely accepted Igbo code-switches. For example: telecommunications - > telekomunikashions, telekomunikashuns or senate -> sineeti, sinetị.

So at first I began to painstakingly edit the `ig.txt` file for about an hour. Realizing the infeasibility of it, I used `max-frequency = 1` and reviewed the output. This way I got **2142** sentences which are correct. Formal review from others will follow shortly.

src/rules/ig.toml
MichaelKohler: I'm mostly offline the next few days, maybe @stefangrotz could help out here. However I don't see this as a necessity per se, if the quality is good enough.

stefangrotz: Hey, what problems do you have with adding these letters @chrisemezue? It should be able to simply add them all. 

For Esperanto I simply collected all letters (uppercase and lowercase) of the alphabet, and it worked, this might be the easiest solution for you:

`allowed_symbols_regex="[AaBbCcDdEeFfGgHhIiJjKkLlMmNnOoPpRrSsTtUuVvZzĉĈĝĜĥĤĵĴŝŜŭŬ;‚'–\\. \"\\?!«»()]"`

stefangrotz: I see. If you want to find out which symbols you want to filter, I can recommend using for example the word count file and then use 

`sed 's/\(.\)/\1\n/g' your-text.txt | sort | uniq -ic > characters-count.txt`

This will give you a file where you easily can collect unwanted signs for the filter.

chrisemezue: thank you for this line of code. very useful.

src/rules/ig.toml
MichaelKohler: If I had to guess, without knowing the language, I'd have the following input:

* The `|` within`Éʻ|c` might not be what you want?
* `\\.` within `\\.pqYð` should already be covered with `;‚'–\\. \"`
* `\”` doesn't need the `\` before, this leads to the one of the parse errors
* Once that is fixed, there is another error complaining about what you just posted while I was writing this. Happy you fixed it :)
* `;` and `,` might also be covered twice (didn't cross-check the actual code points though)

MichaelKohler: There are some other duplications as well, such as `LER`. Might be worth cross-checking those and simplifying. Generally I would say, while it's still quite hard to read, it's still better than the disallow list. Nice work so far!

MichaelKohler: It not being included might be due to the 3 sentences per article requirement and it just picked 3 other sentences from that article? You could pass `--no_check` at least for the validation (but that won't be used for the extract, as the 3 sentences per article rule is a strict limitation). With the `--no_check` flag it basically uses all sentences.

MichaelKohler: Oh no, it actually doesn't match the `w` in that sentence as it's not allowed.

chrisemezue: i fixed the `w` error. it still did not show.

MichaelKohler: Then it might also be the 3 sentences per article limit. I think for verification of the rules it would make sense to use `--no_check` flag when running the extractor.

chrisemezue: you're right. 
everything works well now.

MichaelKohler: Sure, I'm not really concerned about performance here. I'd be fine with this approach. Please report back once the final review is done and I will take it from there. Thanks!

MichaelKohler (approved):

PR #162, opened 2021-10-25
Improvement for best practices for EN rules file (fixes #156)

### LINKED ISSUE #156: Use best practice in English rules file
Over the past years there were quite a few changes to the possible rule configurations. The English rules file still uses some older rules that should now be replaced by better approaches, such as replacing `disallowed_symbols` with `allowed_symbols`.

As most new rule files get copied from EN, this would improve overall quality quite substantially.

MichaelKohler:
This uses best practices learned over the past 2 years or so. As English often gets copied, we for example want to use allowed_symbols.

See #156 for the discussions around this.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -458,9 +458,9 @@ mod test {
     #[test]
     fn test_english() {
         let rules : Rules = load_rules("en");
-        
+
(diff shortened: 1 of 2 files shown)


MichaelKohler:
Review is looking good, I'm merging this and will create a re-run for English. Thanks for the help here!

PR #169, opened 2022-01-08
Add Bengali bn

arijitx:
Basic Rules for Bengali, updated segmenter with nltk PunktSentenceTokenizer as punkt doesnt have a pretrained model in nltk for bengali so used the sentence end as '।'

How many sentences did you get at the end?
~ 129000

How did you create the blacklist file?
Followed instruction in readme, chose words with frequency 50 or less

Evaluate random 500 sentences ?
[500 random sentences](https://docs.google.com/spreadsheets/d/1BJFLOq3L1gDYnHfMVSfUH_r43HS2HspDfrT0Gv3DREw/edit?usp=sharing)
I do not know any linguist any Bengali speaker please feel free to comment on the error ratio. My analysis shows very less number of errors in the extracted sentences.


MichaelKohler:
Thanks for working on this. I will have a few comments, will have a closer look later tonight or tomorrow.

Regarding linguists, would be great, but it's fine if it's "just" native speakers. For the amount, it should be more than 100 sentences though. I would suggest 500. Once I have added my comments, there is some work to be done, so I would suggest to wait with the review until then.

arijitx:
Sure @MichaelKohler waiting for your comments

src/rules/bn.toml
```diff
@@ -0,0 +1,39 @@
+
+# Replacements which are run before sentences are split
+replacements = []
+
+min_trimmed_length = 3
+min_word_count = 3
```
MichaelKohler, line 6: Unlikely, but just to make sure: there are no useful sentences with 2 or less words in Wikipedia. Right?

src/segmenter.rs
```diff
@@ -32,6 +33,7 @@ pub fn split_sentences_with_python_en(text: &str) -> Vec<String> {
     ctx.get("split_sentences")
 }
 
+
```
MichaelKohler, line 36: minor: please remove this new linebreak

src/segmenter.rs
```diff
@@ -70,3 +101,5 @@ mod test {
         assert_eq!(split_sentences_with_python(language, text).len(), 2);
     }
 }
+
+
```
MichaelKohler, line 105: minor: please remove this new linebreak

MichaelKohler (changes requested):
Thanks again! I've added a few remarks and questions :)

src/rules/bn.toml
arijitx: It's rare, tough there can be meaningful phrases as well as entities which can make sense. I will try out with 2 min words and check if that add reasonable words in the data.

src/rules/bn.toml
```diff
@@ -0,0 +1,39 @@
...
+]
+
+even_symbols = ['"']
+
+matching_symbols = [
+  ["„", "“"],
+  ["(", ")"]
+]
+
+abbreviation_patterns = [
```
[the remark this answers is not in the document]
arijitx: Updated, I thought as most abbr. in Bengali phonetic in nature, I will update with some rules following en rules.

src/rules/bn.toml
```diff
@@ -0,0 +1,39 @@
...
+max_word_count = 20
+min_characters = 5
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = false
+broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;"]
+
+disallowed_symbols = [
```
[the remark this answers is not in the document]
arijitx: Thanks for the suggestion, yes, it sounds a better way to add with allowed_symbols_regex.

src/rules/bn.toml
```diff
@@ -0,0 +1,52 @@
...
+may_end_with_colon = false
+quote_start_with_letter = true
+needs_punctuation_end = false
+needs_letter_start = true
+needs_uppercase_start = false
+broken_whitespace = ["  ", " ,", " .", " ?", " !", " ;"]
+
+allowed_symbols_regex = "[॰ঁংঃঅআইঈউঊঋঌএঐওঔকখগঘঙচছজঝঞটঠডঢণতথদধনপফবভমযরলশষসহ়ঽািীুূৃৄেৈোৌ্ৎৗয়ৠৡ৤০১২৩৪৫৬৭৮৯ৰৱ৳৹৺ਂ!\"',-\\./:;\\?`।„“‚‘’ ]"
+
+disallowed_symbols = [
```
MichaelKohler, line 28: This can now be completely removed

src/rules/bn.toml
```diff
@@ -0,0 +1,52 @@
...
+
+even_symbols = ['"']
+
+matching_symbols = [
+  ["„", "“"],
+  ["(", ")"]
+]
+
+abbreviation_patterns = [
+  "([\u0995-\u09B9\u09CE\u09DC-\u09DF\u0985-\u0994\u09BE-\u09CC\u09D7\u09BC]+\\.+)+[\u0995-\u09B9\u09CE\u09DC-\u09DF\u0985-\u0994\u09BE-\u09CC\u09D7\u09BC]*"
```
MichaelKohler, line 48: Could you add a comment to better explain what it does? I think that would be useful for whoever looks at it in the future. Example here: https://github.com/common-voice/cv-sentence-extractor/blob/main/src/rules/en.toml#L36

MichaelKohler (changes requested):
Two more minor comments, the rest looks good to me.

arijitx:
Hi @MichaelKohler I have applied your suggest changes, please feel free to review, I have also created another set of 500 sentences with the updated rules. Do let me know if am good to start the review.

src/rules/bn.toml
arijitx: Is it okay to keep digits in the allowed_symbols_regex ?

src/rules/bn.toml
arijitx: Thanks for suggestion, have added some comments.

src/rules/bn.toml
MichaelKohler: Oh, good question! We don't allow numbers in general, as it's not necessarily clear how to pronounce them. 2022 could for example be "two-zero-two-two", "twenty-twenty-two" or "two thousand and twenty two". Is that the same for Bengali? If so, let's not allow numbers.

arijitx: Got it my bad. I will update the rules to exclude numbers

arijitx:
Thanks for your quick reviews @MichaelKohler do let me know if am good to start with the error rate review process ?

src/segmenter.rs
```diff
@@ -61,6 +84,13 @@ mod test {
         assert_eq!(split_sentences_with_python(language, text).len(), 2);
     }
 
+    fn test_segmenter_bn() {
```
MichaelKohler, line 87:
```suggestion
    #[test]
    fn test_segmenter_bn() {
```

MichaelKohler (changes requested):
Thanks for the changes! This looks good apart from one more comment I missed before. However this does not block any review. At the bottom of https://github.com/common-voice/cv-sentence-extractor/actions/runs/1674625643 you find a "extraction" artifact. That zip file contains a `sample.txt` which you can use for the review. I would suggest to shuffle this file randomly, to make sure that extractions are not all from the same article (`shuf sample.txt > shuffled.txt`).

arijitx:
Thanks @MichaelKohler , I have downloaded the sample and shuffled it randomly and taken 500 for further review.

MichaelKohler (approved):
Looks good to me! Approving the code changes and looking forward to the result of the review!

arijitx:
Hi @MichaelKohler,

I could arrange two reviews, here are the results still WIP for the 3rd review

### Review 1:
**Correct rate: 95%**
https://docs.google.com/spreadsheets/d/1BJFLOq3L1gDYnHfMVSfUH_r43HS2HspDfrT0Gv3DREw/edit?usp=sharing

### Review 2:
**Correct rate 96%**
https://docs.google.com/spreadsheets/d/1XPjdZzyzqrXSCwTJvAtD6sK3L-00i5CERz-QBTk0MfQ/edit?usp=sharing

MichaelKohler:
Now that you have the reviews, did ypu notice a pattern in the wrong sentences? Anything that easily could be filtered out? If so, would be great to add that and do another review. If not, also ok :)

arijitx:
Hi @MichaelKohler , From my analysis most of the error is like incomplete/partial sentence or phrases, there isn't a very high number of them, either its written that way and probably from the context of article it made sense. I don't see a clear way to filter such sentences.

MichaelKohler (approved):
Looking all good then. Thanks for working on this! I will start the extract and CC you on the Common Voice PR I'm gonna create.

PR #177, opened 2022-08-10
Add more abbreviations for DE detected when running a re-export

MichaelKohler:
I re-ran a DE wikipedia extract and found these abbreviations that we can catch.

Review: https://docs.google.com/spreadsheets/d/1UYkov-e6KlHAXOI2o5nnY5xqV52B1I156MDsywi6Tik/edit#gid=0

@stefangrotz could you have a quick look at this? I will do a quick review as well, I think for this case one reviewer will be okay, but it would be nice if you could have a general look at the additions and tell me if we shouldn't do something.

### DIFF
--- a/src/rules/de.toml
+++ b/src/rules/de.toml
@@ -23,6 +23,7 @@ replacements = [
   ["mind.", "mindestens"],
   ["evtl.", "eventuell"],
   ["bzgl.", "bezüglich"],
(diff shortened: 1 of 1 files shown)


MichaelKohler:
@stefangrotz any objections to this from your side? If not, I'll merge this and then run a re-run.

stefangrotz:
Looks good to me :)

PR #179, opened 2022-10-20
Update de.toml

BrunoFischerGermany:
I have added new abbreviations.
rd - rund
hl - heilig
dgl - der gleichen
No. - Number
Abk - Abkürzung
Abb. - Abbildung
geb. - geboren
gegr. - gegründet

### DIFF
--- a/src/rules/de.toml
+++ b/src/rules/de.toml
@@ -32,6 +32,8 @@ replacements = [
   ["Abb.", "Abbildung"],
   ["geb.", "geboren"], 
   ["gegr.", "gegründet"],
+  ["vs.", "versus"],
+  ["kgl", "königlich"],
 ]
 
 segmenter = "python"
--- a/src/rules/de.toml
+++ b/src/rules/de.toml
@@ -24,6 +24,14 @@ replacements = [
   ["evtl.", "eventuell"],
   ["bzgl.", "bezüglich"],
   ["Nr.", "Nummer"],
+  ["rd.", "rund"], 
+  ["hl.", "heilig"],
+  ["dgl.", "Der gleichen"],
+  ["No.", "Number"],
+  ["Abk.", "Abkürzung"], 
+  ["Abb.", "Abbildung"],
+  ["geb.", "geboren"], 
+  ["gegr.", "gegründet"],
 ]
 
 segmenter = "python"


stefangrotz:
Danke für den Pull Request! 

dgl würde ich zu "dergleichen" statt "Der gleichen" machen.

@MichaelKohler  fällt Dir noch etwas auf?

MichaelKohler:
Apart from Stefan's comment, this looks good to me. I'm also gonna have a look at the sample output to make sure this has no unintended side effects once that change is made. 

Thanks!

BrunoFischerGermany:
Ich habe noch 2 Abkürzungen beim Spenden meiner Stimme gefunden.

vs - versus und kgl königlich

Viele Grüße

src/rules/de.toml
```diff
@@ -32,6 +32,8 @@ replacements = [
   ["Abb.", "Abbildung"],
   ["geb.", "geboren"], 
   ["gegr.", "gegründet"],
+  ["vs.", "versus"],
+  ["kgl", "königlich"],
```
MichaelKohler, line 36: Is this really occuring without a period? I wonder how many words there are that have "kgl" in the middle somewhere, which would be replaced as well 😅 nothing comes to mind right now, but might not be safe

BrunoFischerGermany: Oh maybe is it better with kgl. 

I found 

eidg. - Eidgenossenschaftlich

Maybe add this Zoo?

src/rules/de.toml
```diff
@@ -24,6 +24,16 @@ replacements = [
   ["evtl.", "eventuell"],
   ["bzgl.", "bezüglich"],
   ["Nr.", "Nummer"],
+  ["rd.", "rund"], 
+  ["hl.", "heilig"],
```
MichaelKohler, line 28: Just came to my mind, I don't think this one will work nicely. I generally think that adjectives won't work nicely due to the different cases.

"Die hl. drei Könige" -> heiligen
"Die hl. Maria" -> heilige

src/rules/de.toml
```diff
@@ -24,6 +24,16 @@ replacements = [
...
+  ["dgl.", "Der gleichen"],
```
MichaelKohler, line 29:
```suggestion
  ["dgl.", "dergleichen"],
```

src/rules/de.toml
MichaelKohler: See my comment above, same applies here. And I think the same would also apply for "eidg.". What do you think?

MichaelKohler (changes requested):
Sorry for the delay here. Added a few comments.

src/rules/de.toml
BrunoFischerGermany: > Is this really occuring without a period? I wonder how many words there are that have "kgl" in the middle somewhere, which would be replaced as well 😅 nothing comes to mind right now, but might not be safe

you are right. Not without A Period.

So "kgl." And "Eidg." Are better. 👌

Issue #182, opened 2023-06-13
max_characters in rules


HarikalarKutusu:
I'm trying to make this work with Turkish.
I see the following:
```toml
min_trimmed_length = 
min_word_count = 
max_word_count = 
min_characters = 
```
If I'm not mistaken, there is no max_characters setting. Like in German, Turkish words have a high variance in length due to the agglutinative nature of the language. So, a 5-6 word sentence can be quite long while reading.

I've been also planning to change the sentence-collector rules to use charter length instead of words, but I can see that it is missing here.

If this is true, can this be added?

MichaelKohler:
I think that sounds reasonable to be implemented. I'd suggest to implement this the same way as `min_characters`, including a test for it and documentation in the README.

Thanks for bringing this up.

HarikalarKutusu:
OK, I'll do it tomorrow on a clean clone (I've been messing with the current one)...

HarikalarKutusu:
Implemented with #183

PR #183, opened 2023-06-13
Add max_characters to rules

HarikalarKutusu:
Implements #182

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -3,6 +3,11 @@ use toml::Value;
 use regex::Regex;
 
 pub fn check(rules: &Rules, raw: &str) -> bool {
+
+    fn in_limit(x: usize, min_val: usize, max_val: usize) -> bool {
+        return (x >= min_val) && (x <= max_val)
+    }
+
     let trimmed = raw.trim();
     if trimmed.len() < rules.min_trimmed_length
         || rules.quote_start_with_letter
@@ -12,7 +17,7 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
                 .nth(1)
                 .map(|c| !c.is_alphabetic())
                 .unwrap_or_default()
-        || trimmed.chars().filter(|c| c.is_alphabetic()).count() < rules.min_characters
+        || !in_limit(trimmed.chars().filter(|c| c.is_alphabetic()).count(), rules.min_characters, rules.max_characters)
         || !rules.may_end_with_colon && trimmed.ends_with(':')
         || rules.needs_punctuation_end && trimmed.ends_with(|c: char| c.is_alphabetic())
         || rules.needs_letter_start && trimmed.starts_with(|c: char| !c.is_alphabetic())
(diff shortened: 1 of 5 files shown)


src/checker.rs
```diff
@@ -3,6 +3,11 @@ use toml::Value;
 use regex::Regex;
 
 pub fn check(rules: &Rules, raw: &str) -> bool {
+
+    fn in_limit(x: usize, min_val: usize, max_val: usize) -> bool {
+        return (x >= min_val) && (x <= max_val)
```
MichaelKohler, line 8: Minor: could you move this outside the function?

src/rules.rs
```diff
@@ -66,6 +67,7 @@ impl Default for Rules {
             min_word_count: 1,
             max_word_count: 14,
             min_characters: 0,
+            max_characters: std::usize::MAX,
```
MichaelKohler, line 70: `std::usize::MAX` is deprecated, could you replace it with `usize::MAX`? https://doc.rust-lang.org/std/primitive.usize.html#associatedconstant.MAX

src/rules.rs
```diff
@@ -100,6 +102,7 @@ mod test {
         assert_eq!(rules.min_word_count, 1);
         assert_eq!(rules.max_word_count, 14);
         assert_eq!(rules.min_characters, 0);
+        assert_eq!(rules.max_characters, 999);
```
MichaelKohler, line 105: Where is 999 coming from? I think the tests failed before so you didn't get an automated comment here. This should be `usize::MAX`.

src/checker.rs
```diff
@@ -12,7 +17,7 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
                 .nth(1)
                 .map(|c| !c.is_alphabetic())
                 .unwrap_or_default()
-        || trimmed.chars().filter(|c| c.is_alphabetic()).count() < rules.min_characters
+        || !in_limit(trimmed.chars().filter(|c| c.is_alphabetic()).count(), rules.min_characters, rules.max_characters)
```
MichaelKohler, line 20: To make this line a bit shorter, what do you think about assigning the count to a variable before this fill check?

src/checker.rs
```diff
@@ -139,6 +144,17 @@ mod test {
...
 
+    #[test]
+    fn test_max_characters() {
+        let rules : Rules = Rules {
+            max_characters: 25,
+            ..Default::default()
+        };
+
+        assert!(!check(&rules, &String::from("not less than or equal to 25")));
+        assert!(check(&rules, &String::from("less than or equal to 25")));
```
MichaelKohler, line 155: I'm not seeing anything obviously wrong here, can you have a deeper look into why this test case fails? Might also be some default rule?

MichaelKohler (changes requested):
Thank you! Generally this looks good to me, I've left a few comments.

src/rules.rs
HarikalarKutusu: Yep just saw that and already replaced it everywhere.

src/checker.rs
HarikalarKutusu: Yep, better...

src/checker.rs
HarikalarKutusu: I'm not sure either. I was playing with MAX and 999 but could not pass the tests yet (my first lines of code in Rust).

src/rules.rs
HarikalarKutusu: Done...

src/checker.rs
HarikalarKutusu: Done...

src/checker.rs
HarikalarKutusu: Let's see if it is fixed here also.

MichaelKohler (approved):
Thanks for the other `usize::MAX` adjustments. All looking good to me now. Thanks, well done on your first Rust contribution!

PR #187, opened 2023-06-24
Add a new rule - stem_separator_regex

HarikalarKutusu:
This will mainly be useful for blacklisted proper names with suffixes. If you blacklist the stem word (e.g. a person's name) it should be enough.

If specified, the code splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. 

It is a simple regex of separators. For example, for apostrophes, you specify `stem_separator_regex = "[']"` in the rule file.

If you do not specify it, or set it to `= ""` or `= "[]"` it will not be triggered.

It works after the initial blacklist check is done and only checks stem words extracted with stem_separator_regex against the blacklist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,7 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_punctuation_end |  If a sentence needs to end with a punctuation | boolean | false
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
+| stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
 | quote_start_with_letter |  If a quote needs to start with a letter | boolean | true
 | replacements |  Replaces abbreviations or other words according to configuration. This happens before any other rules are checked. | Array of replacement configurations: each configuration is an Array of two values: `["search", "replacement"]`. See example below. | nothing gets replaced
 | segmenter |  Segmenter to use for this language. See below for more information. | "python" | using `rust-punkt` by default
--- a/README.md
+++ b/README.md
@@ -174,10 +174,10 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_punctuation_end |  If a sentence needs to end with a punctuation | boolean | false
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
-| stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
 | quote_start_with_letter |  If a quote needs to start with a letter | boolean | true
 | replacements |  Replaces abbreviations or other words according to configuration. This happens before any other rules are checked. | Array of replacement configurations: each configuration is an Array of two values: `["search", "replacement"]`. See example below. | nothing gets replaced
 | segmenter |  Segmenter to use for this language. See below for more information. | "python" | using `rust-punkt` by default
+| stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
 
 ### Example for `matching_symbols`
 
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -63,9 +63,9 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
         let mut stems_words: Vec<&str> = vec![];
         
         for word in words {
-            let maybe_stem_word: &str = &(regex.split(word).nth(0).unwrap_or(word));
+            let maybe_stem_word: &str = regex.split(word).next().unwrap_or(word);
             if maybe_stem_word != word {
-                stems_words.push(&maybe_stem_word);
+                stems_words.push(maybe_stem_word);
             }
         }
 
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -58,19 +58,18 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
         return false;
     }
 
-    if !rules.stem_separator_regex.is_empty() && !(rules.stem_separator_regex == "[]") {
+    if !rules.stem_separator_regex.is_empty() {
         let regex: Regex = Regex::new(&rules.stem_separator_regex).unwrap();
         let mut stems_words: Vec<&str> = vec![];
         
         for word in words {
-            let maybe_stem_word: &str;
-            maybe_stem_word = &(regex.split(word).nth(0).unwrap_or(word));
+            let maybe_stem_word: &str = &(regex.split(word).nth(0).unwrap_or(word));
             if maybe_stem_word != word {
                 stems_words.push(&maybe_stem_word);
             }
         }
 
-        if stems_words.clone().into_iter().any( |word| rules.disallowed_words.contains(word))
+        if stems_words.into_iter().any(|word| rules.disallowed_words.contains(word))
         {
             return false;
         }
@@ -333,11 +332,10 @@ mod test {
         assert!(!check(&rules, &String::from("Washington DC's Mall has many musums.")));
 
         let rules : Rules = Rules {
-            stem_separator_regex: "[]".to_string(),
             disallowed_words: ["Smithsonian", "DC", "Museum"].iter().map(|s| (*s).to_string()).collect(),
             ..Default::default()
         };
-        assert!(check(&rules, &String::from("Smithsonian's venues are in DC's Mall will not be checked for stem words.")));
+        assert!(check(&rules, &String::from("Smithsonian's venues are in DC's Mall - this will not be checked for stem words.")));
 
 
     }
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -63,14 +63,13 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
         let mut stems_words: Vec<&str> = vec![];
         
         for word in words {
-            let maybe_stem_word: &str = regex.split(word).next().unwrap_or(word);
+            let maybe_stem_word = regex.split(word).next().unwrap_or(word);
             if maybe_stem_word != word {
                 stems_words.push(maybe_stem_word);
             }
         }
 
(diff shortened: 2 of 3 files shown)


MichaelKohler:
Thanks for the PR, I will have a look at it in the next few days.

src/checker.rs
```diff
@@ -47,17 +47,35 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
         return false;
     }
 
-    let mut words = trimmed.split_whitespace();
+    let words = trimmed.split_whitespace();
```
MichaelKohler, line 50: I'd suspect that the reason this was mutable was for the `words.any()` call to work further down. However I don't think this is the right fix for the underlying problem, so I prefer getting rid of this `mut` here.

While cloning `words` twice needs some memory, this is nowhere close to being an actual problem with the usual sizes of a Wikipedia article or anything else that could be fed to `sentence-extractor`. Of course in theory you could feed it gigabytes of data through a single file, but I'd argue we'd still be fine here.

src/checker.rs
```diff
@@ -47,17 +47,35 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
...
         || word_count > rules.max_word_count
-        || words.any(|word| rules.disallowed_words.contains(
+        || words.clone().any(|word| rules.disallowed_words.contains(
              &word.trim_matches(|c: char| !c.is_alphabetic()).to_lowercase()
            ))
     {
         return false;
     }
 
+    if !rules.stem_separator_regex.is_empty() && !(rules.stem_separator_regex == "[]") {
```
MichaelKohler, line 61:
```suggestion
    if !rules.stem_separator_regex.is_empty() && rules.stem_separator_regex != "[]" {
```

That being said, I don't think we should specifically check for `[]`. That is invalid is per:

```
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
regex parse error:
    []
    ^^
error: unclosed character class
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
)', src/checker.rs:62:68
```

and IMHO should lead to an actual error instead of us swallowing that case silently. The same goes for any other invalid regex. WDYT?

src/checker.rs
```diff
@@ -47,17 +47,35 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
...
+        let regex: Regex = Regex::new(&rules.stem_separator_regex).unwrap();
+        let mut stems_words: Vec<&str> = vec![];
+        
+        for word in words {
+            let maybe_stem_word: &str;
+            maybe_stem_word = &(regex.split(word).nth(0).unwrap_or(word));
```
MichaelKohler, line 67: together with the line above:

```suggestion
            let maybe_stem_word = regex.split(word).next().unwrap_or(word);
```

src/checker.rs
```diff
@@ -47,17 +47,35 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
...
+            if maybe_stem_word != word {
+                stems_words.push(&maybe_stem_word);
+            }
+        }
+
+        if stems_words.clone().into_iter().any( |word| rules.disallowed_words.contains(word))
```
MichaelKohler, line 73: This `clone()` is not needed.

MichaelKohler, line 73: Could you please remove the space between `(` and `|word|`? And then also move the function `{` on to the same line? That would fit our pattern better, even though we do not 100% use that pattern (though we should).

src/checker.rs
```diff
@@ -301,6 +319,29 @@ mod test {
...
+            ..Default::default()
+        };
+
+        assert!(check(&rules, &String::from("The Mall has many museums.")));
+        assert!(!check(&rules, &String::from("Smithsonian's venues are in the Mall.")));
+        assert!(!check(&rules, &String::from("Do you know Smithsonian's African American Museum's location?")));
+        assert!(!check(&rules, &String::from("Washington DC's Mall has many musums.")));
+
+        let rules : Rules = Rules {
+            stem_separator_regex: "[]".to_string(),
```
MichaelKohler, line 336: This is an invalid regex, can you replace this test case with a case that tests if it's not provided at all?

README.md
```diff
@@ -174,6 +174,7 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_punctuation_end |  If a sentence needs to end with a punctuation | boolean | false
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
+| stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
```
MichaelKohler, line 177: Could you please sort this alphabetically? (at the bottom of the table)

MichaelKohler (changes requested):
Thanks for this PR, this is generally looking good to me, I just have some minor comments.

src/checker.rs
HarikalarKutusu: In the original "words" was only used once, so that was not a problem. But with this PR I use it again. So it became a must to make it immutable and clone it as `words.clone().any()`.

AFAIK, the check function is called per sentence, and a sentence should fit 10 sec as of now, so it will not be huge.

src/checker.rs
HarikalarKutusu: I added it to hide user errors for better UX. People might set it to [] and might not know what the error means.

The whole experience is based on experimentation, adding/removing etc. Such user mistakes might happen.

But it is your decision...

src/checker.rs
HarikalarKutusu: Right :) As you know, I had to divide the one-liners to see what is happening in the debugger.

src/checker.rs
HarikalarKutusu: Right, again leftover from trying to fix a bug.

src/checker.rs
MichaelKohler: > AFAIK, the check function is called per sentence, and a sentence should fit 10 sec as of now, so it will not be huge.

Correct, you're right. Even less likely a problem then.

src/checker.rs
HarikalarKutusu: Done.

src/checker.rs
HarikalarKutusu: Done.

src/checker.rs
HarikalarKutusu: Done.

src/checker.rs
HarikalarKutusu: OK, done.

README.md
HarikalarKutusu: Oh, I wasn't aware of that. Sorry.
Done.

src/checker.rs
HarikalarKutusu: Removed that case as suggested.

src/checker.rs
```diff
@@ -301,6 +318,28 @@ mod test {
...
+        assert!(check(&rules, &String::from("The Mall has many museums.")));
+        assert!(!check(&rules, &String::from("Smithsonian's venues are in the Mall.")));
+        assert!(!check(&rules, &String::from("Do you know Smithsonian's African American Museum's location?")));
+        assert!(!check(&rules, &String::from("Washington DC's Mall has many musums.")));
+
+        let rules : Rules = Rules {
+            disallowed_words: ["Smithsonian", "DC", "Museum"].iter().map(|s| (*s).to_string()).collect(),
+            ..Default::default()
+        };
+        assert!(check(&rules, &String::from("Smithsonian's venues are in DC's Mall - this will not be checked for stem words.")));
```
MichaelKohler, line 338: Could you please have a look at this test failure?

src/checker.rs
```diff
@@ -301,6 +318,28 @@ mod test {
...
+
+
```
MichaelKohler, line 340: Oh, didn't write on this before, can you please also remove those two empty lines?

src/checker.rs
```diff
@@ -47,17 +47,34 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
...
+        
+        for word in words {
+            let maybe_stem_word: &str = &(regex.split(word).nth(0).unwrap_or(word));
+            if maybe_stem_word != word {
+                stems_words.push(&maybe_stem_word);
+            }
+        }
+
+        if stems_words.into_iter().any(|word| rules.disallowed_words.contains(word))
+        {
```
MichaelKohler, line 73: very minor and really only because you will need to adjust it anyway: can you move the `{` brace to the if line as well?

src/checker.rs
```diff
@@ -47,17 +47,34 @@ pub fn check(rules: &Rules, raw: &str) -> bool {
...
     {
         return false;
     }
 
+    if !rules.stem_separator_regex.is_empty() {
+        let regex: Regex = Regex::new(&rules.stem_separator_regex).unwrap();
+        let mut stems_words: Vec<&str> = vec![];
+        
+        for word in words {
+            let maybe_stem_word: &str = &(regex.split(word).nth(0).unwrap_or(word));
```
MichaelKohler: I think you missed two other things that were in my previous comment, though I should have mentioned them explicitly as well.

```suggestion
            let maybe_stem_word = regex.split(word).next().unwrap_or(word);
```

* The reference is not needed here
* As `.nth(0)` and `.next()` are the same, let's be more idiomatic and use `.next()`

MichaelKohler (changes requested):
Thanks for the corrections, just two more things and a test failure. Then I think we can merge this PR :)

src/checker.rs
HarikalarKutusu: I could not figure it out :(

src/checker.rs
HarikalarKutusu: Missed the next() and reference removal before... It was also becoming && on the next line...
This way is the last iteration:
````
        for word in words {
            let maybe_stem_word = regex.split(word).next().unwrap_or(word);
            if maybe_stem_word != word {
                stems_words.push(maybe_stem_word);
            }
        }
```

src/checker.rs
HarikalarKutusu: Man, it was longer than 14 words :(
Finally got it.

MichaelKohler (approved):
Thank you!

MichaelKohler:
This will mainly be useful for blacklisted proper names with suffixes. If you blacklist the stem word (e.g. a person's name) it should be enough.

If specified, the code splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist.

It is a simple regex of separators. For example, for apostrophes, you specify stem_separator_regex = "[']" in the rule file.

If you do not specify it, it will not be triggered.

HarikalarKutusu:
Actually, thank YOU! It took too much of your time but resolved a (for me) major issue.

PR #194, opened 2023-06-29
Fix: stem_separator_regex does not work on real data

HarikalarKutusu:
And it is because, in the real world, the `rules.disallowed_words` contains lowercase words. The tests were working without any conversion.
This fix handles both the main code and the tests by making them lowercase.


MichaelKohler (approved):
Thanks!

Issue #198, opened 2023-07-08
Feature Request - Removal of parentheses.


HarikalarKutusu:
Parentheses: They are problematic for the Common Voice use.
1. They are disruptive to the natural speech flow, they are not used in everyday speech. If read from a text, they can be "read" such as "open parenthesis" and "parenthesis closed" which will invalidate the recordings in CV.
2. Wiki resources many times use parentheses to refer to original wording in the original language/how it is pronounced etc. These are mostly in another alphabet and their existence removes the sentence from possible sentences.

Here are some examples for the second point - ref. article [Istanbul](https://en.wikipedia.org/wiki/Istanbul):
```
The city was founded as Byzantium (Greek: Βυζάντιον, Byzantion) in the 7th century BCE by Greek settlers from Megara.

In English the stress is on the first or last syllable, but in Turkish it is on the second syllable (-tan-).

On the European side, near the point of the peninsula (Sarayburnu), there was a Thracian settlement during the early 1st millennium BCE.
```

If you allow them in the `allowed_symbols_regex`, the first problem will arise.

They should not be in the CV text corpora, and removal of them would increase getting more possibilities for the random selection (rules related + shortened sentences + lesser known words blacklisted, etc), especially important for lower resource languages. I expect a 10-20% increase in the raw resource (to be measured).

So I propose (and start to implement) a rule "remove_parentheses", boolean, default false.

HarikalarKutusu:
This one is handled by #199 with a more general approach, namely the use of a user-defined bracket list.

PR #199, opened 2023-07-08
Add rule to remove content between brackets

HarikalarKutusu:
This PR adds a new rule to optionally remove brackets from sentences before they are checked for other rules.

It will handle the issues mentioned in #198 ...

It will work like "replaces", before replaces and before any rule-check, so you can also run it with --no_check. This way you get all sentences but with content in parentheses removed.

You specify opening and closing brackets as a list, and they are executed from top to bottom.
```
[
  ["(", ")"],
  ["[", "]"],
]
```

- It can handle these cases:
  - This (definitely) is removed. => This is removed.
  - Works with (first) multiple (second) parentheses. => Works with multiple parentheses .
  - Handles nested (outer and [inner] and (another inner)) brackets => Handles nested brackets.

- Edge cases which cannot be handled:
  - Non-matching ((huh) brackets at the source. => Non-matching ( brackets at the source.
  - If not surrounded by spaces space is inserted (here). => If not surrounded by spaces space is inserted .

Tested on Turkish Wikipedia, with the same rules turning this feature off and on (no blacklist & no export limitation - i.e. 3 sentence/article limit removed): 5.8% increase in the number of total possible sentences (947,526 => 1,002,522)

EDITS:
- Changed this post to reflect the final code
- TESTED ON REAL DATA, CAN BE MERGED.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
 | quote_start_with_letter |  If a quote needs to start with a letter | boolean | true
-| remove_brackets_list |  Removes (possibly nested) user defined brackets and content inside them `(anything [else])` from the sentence before checking other rules | Array of matching brackets: each configuration is an Array of two values: `["opening_bracket", "closing_bracket"]`. See example below. | []
+| remove_brackets_list |  Removes (possibly nested) user defined brackets and content inside them `(anything [else])` from the sentence before replacements and checking other rules | Array of matching brackets: each configuration is an Array of two values: `["opening_bracket", "closing_bracket"]`. See example below. | []
 | replacements |  Replaces abbreviations or other words according to configuration. This happens before any other rules are checked. | Array of replacement configurations: each configuration is an Array of two values: `["search", "replacement"]`. See example below. | nothing gets replaced
 | segmenter |  Segmenter to use for this language. See below for more information. | "python" | using `rust-punkt` by default
 | stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
@@ -209,7 +209,7 @@ remove_brackets_list = [
 ]
 ```
 
-This internally creates regex patterns `(.*)` and `[.*]` and removes them with their content, executed in given order.
+This internally creates related regex patterns and removes them with their content, executed in given order.
 
 ```
 Input: This (parantheses) (and this) will be removed also this one (another [one]) should.
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
 | quote_start_with_letter |  If a quote needs to start with a letter | boolean | true
-| remove_parentheses |  Removes (possibly nested) parentheses and content inside them `(anything)` from the sentence before checking other rules | boolean | false
+| remove_brackets_list |  Removes (possibly nested) user defined brackets and content inside them `(anything [else])` from the sentence before checking other rules | Array of matching brackets: each configuration is an Array of two values: `["opening_bracket", "closing_bracket"]`. See example below. | []
 | replacements |  Replaces abbreviations or other words according to configuration. This happens before any other rules are checked. | Array of replacement configurations: each configuration is an Array of two values: `["search", "replacement"]`. See example below. | nothing gets replaced
 | segmenter |  Segmenter to use for this language. See below for more information. | "python" | using `rust-punkt` by default
 | stem_separator_regex |  If given, splits words at the given characters to reach the stem words to check them again against the blacklist, e.g. prevents "Rust's" to pass if "Rust" is in the blacklist. | Simple regex of separators, e.g. for apostrophe `stem_separator_regex = "[']"` | ""
@@ -198,6 +198,25 @@ Output: Valid
 
 Input: This is (a test))
 Output: Invalid
+```
+
+### Example for `remove_brackets_list`
+
+```
+remove_brackets_list = [
+  ["(", ")"],
+  ["[", "]"]
+]
+```
+
+This internally creates regex patterns `(.*)` and `[.*]` and removes them with their content, executed in given order.
+
+```
+Input: This (parantheses) (and this) will be removed also this one (another [one]) should.
+Output: This will be removed also this one should.
+
+Input: This is (malformed)) at the source.
+Output: This is ) at the source.
 ```
 
 ### Example for `replacements`
--- a/README.md
+++ b/README.md
@@ -175,6 +175,7 @@ The following rules can be configured per language. Add a `<language>.toml` file
 | needs_uppercase_start |  If a sentence needs to start with an uppercase | boolean | false
 | other_patterns |  Regex to disallow anything else | Rust Regex Array | all other patterns allowed
 | quote_start_with_letter |  If a quote needs to start with a letter | boolean | true
+| remove_parentheses |  Removes (possibly nested) parentheses and content inside them `(anything)` from the sentence before checking other rules | boolean | false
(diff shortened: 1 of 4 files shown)


MichaelKohler:
Sorry for the delay here and thanks for submitting this PR. I hope I can get some time in the next few days to review.

HarikalarKutusu:
As specified above, there are two problems with this implementation:
1. It does not take care of nested parentheses, it only handles the inner one. One could add a second regex call, but it might be an unnecessarily costly one.
2. This might result in what you call "broken whitespace" cases if the parentheses are not surrounded by spaces.

Any solution you can find during your examination is very much welcome.

src/checker.rs
```diff
@@ -7,7 +7,9 @@ fn in_limit(x: usize, min_val: usize, max_val: usize) -> bool {
 }
 
 pub fn check(rules: &Rules, raw: &str) -> bool {
-    let trimmed = raw.trim();
+
+    let trimmed: &str = raw.trim();
+
```
MichaelKohler, line 12: Let's remove these two added new-lines here.

src/replacer.rs
```diff
@@ -1,8 +1,18 @@
...
 use toml::Value;
+use regex::Regex;
+
+fn maybe_remove_parentheses(rules: &Rules, txt: &str) -> String {
+    let mut replaced = txt.to_string();
+    if rules.remove_parentheses {
+        let regex = Regex::new("\\([^\\()\n]*\\)").unwrap();
+        replaced = regex.replace_all(txt, "").to_string().replace("  ", " ");
+    }
+    return replaced;
```
MichaelKohler, line 11:
```suggestion
    replaced
```

src/replacer.rs
```diff
@@ -1,8 +1,18 @@
...
+}
 
 pub fn replace_strings(rules: &Rules, raw: &str) -> String {
-    let mut result = raw.trim().to_string();
+    let mut result = maybe_remove_parentheses(&rules, raw.trim());
```
MichaelKohler, line 15:
```suggestion
    let mut result = maybe_remove_parentheses(rules, raw.trim());
```

MichaelKohler (commented):
Thanks again, I had some time to think more about this and have some minor comments on the code as well. Let's first tackle the bigger questions though:

* What's the reason for preventing multi-line matches? While generally we probably shouldn't have new-lines anyway, if we do I probably would expect it to work nevertheless.
* In my opinion these "turn-on-turn-off" style rules should cover the full use case and should not need to be combined with other rules. Therefore I would argue that it should take care of multi paranethesis with eagerly matching them, resulting in a sentence without parenthesis.
* Additionally I would also say that it should take care of whitespaces correctly so it's not needed to combine it with other rules.

That being said, the more I think about it, the more I think it should be a configurable Regex instead of a boolean config. I don't see a good way of deprecating existing configs when we decide to allow them to work for other languages as well. And then maintaining two configs that do almost the same just makes everything more complicated. Given that the Regex is not easy to come up with, we could leave the current config with `false` and `true` (maybe rename it to something more generic). If it's set to `true` the default Regex could be hardcoded. And then we could add a `*_regex` config to overwrite that default. 

I'm on mobile right now so I can provide an example later if needed.

What do you think?

HarikalarKutusu:
Let me try to comment on these first:

> What's the reason for preventing multi-line matches? While generally we probably shouldn't have new-lines anyway, if we do I probably would expect it to work nevertheless.

OK, I missed the following line in the code. I thought we were working on the whole "text" field. Actually the segmenter works on paragraphs. My mistake. So it should not be needed.

https://github.com/common-voice/cv-sentence-extractor/blob/fef72ab03a5e514a8e6d931fdb7f3bc5e6c1164d/src/loaders/wikipedia.rs#L38

But if there were new-lines, I wouldn't like to have too many paragraphs removed because of typos.

> In my opinion these "turn-on-turn-off" style rules should cover the full use case and should not need to be combined with other rules. Therefore I would argue that it should take care of multi paranethesis with eagerly matching them, resulting in a sentence without parenthesis.

One solution would be to run it in a loop and check if the result is the same as the input to break the loop.

> Additionally I would also say that it should take care of whitespaces correctly so it's not needed to combine it with other rules.

We can replace (anything) with space and remove any double spaces with regex afterward.

> maybe rename it to something more generic

A more generic name would be "bracket". I quickly examined the [Wikipedia article](https://en.wikipedia.org/wiki/Bracket). There seem to be a huge number of possibilities (see the Unicode list in the article). The normal parentheses are abundant in Latin script, but I cannot check if the code works in non-Latin languages. Francis also said he did not work with Chinese or others. But as far as I can see, he recently updated commonvoice-utils repo to handle some like Hangul.

This part really needs teamwork - which is why I did not go that way.

A second thing: I'm not sure it can be handled with a simple regex as you mentioned. Probably we would need to feed pairs.

src/replacer.rs
```diff
@@ -127,4 +143,31 @@ mod test {
...
+                Value::try_from([Value::try_from("[").unwrap(), Value::try_from("]").unwrap()]).unwrap(),
+            ],
+            ..Default::default()
+        };
+
+        assert_eq!(replace_strings(&rules, &String::from("Three: (content) should be removed.")), "Three: should be removed.");
+        assert_eq!(replace_strings(&rules, &String::from("Four: [content (and nested one)] should be removed.")), "Four: should be removed.");
+        assert_eq!(replace_strings(&rules, &String::from("Five: (content (and nested one)) should be removed.")), "Five: should be removed.");
+        assert_eq!(replace_strings(&rules, &String::from("Six: (one) (two) and [three] 'and' should stay.")), "Six: and 'and' should stay.");
+        assert_ne!(replace_strings(&rules, &String::from("Seven: Test edge case (here).")), "Seven: Test edge case.");
```
MichaelKohler, line 171: See above, not sure if we really need this.

README.md
```diff
@@ -199,6 +200,25 @@ Input: This is (a test))
...
+### Example for `remove_brackets_list`
+
+```
+remove_brackets_list = [
+  ["(", ")"],
+  ["[", "]"]
+]
+```
+
+This internally creates regex patterns `(.*)` and `[.*]` and removes them with their content, executed in given order.
```
MichaelKohler, line 212: The regex pattern now looks different ;)

src/replacer.rs
```diff
@@ -1,9 +1,25 @@
...
+    // bracket removal
+    for bracket_pair in rules.remove_brackets_list.iter() {
+        if Value::as_array(bracket_pair).unwrap().len() == 2 {
+            let br0 = bracket_pair[0].as_str().unwrap();
+            let br1 = bracket_pair[1].as_str().unwrap();
+            let regex = Regex::new(format!(r#"\{}[^\{}\{}]*\{}"#, br0, br0, br1, br1).as_str()).unwrap();
+            let mut prev = String::from("");
+            while prev != result {
+                prev = result.clone();
+                result = regex.replace_all(prev.as_str(), " ").to_string().replace("  ", " ");
```
MichaelKohler, line 17: While the spaces replacement might not cover all languages, I'm fine with this for now. If needed this can still be adjusted in the future.

MichaelKohler (approved):
Thanks for adjusting this based on the feedback. I think this is a better approach now. Sorry for the delay of reviewing it.

I suppose you have also tested this with your other efforts for Turkish? Did you notice anything off by now?

Overall I think I'm approving this for now, with a few minor comments. Thanks for working on this!

HarikalarKutusu:
> I suppose you have also tested this with your other efforts for Turkish? Did you notice anything off by now?

I tested it with () and [] combined, without any other replacements/rules, and compared the result with original sentences. All matching brackets were removed, and everything else was intact. There were a few sentences out of a million with a single "(" left because the original had non-matching ones, e.g. "This has ((non-matching) brackets".

One point on this: I do not allow these brackets along with my alphabet as I want them removed, so these out-of-control ones get eliminated.

src/checker.rs
```diff
@@ -7,7 +7,7 @@ fn in_limit(x: usize, min_val: usize, max_val: usize) -> bool {
 }
 
 pub fn check(rules: &Rules, raw: &str) -> bool {
-    let trimmed = raw.trim();
+    let trimmed: &str = raw.trim();
```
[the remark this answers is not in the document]
HarikalarKutusu: I like "typing" :) It was needed for type-checking during coding (many String/&str conversions), but I suppose not now.

src/replacer.rs
HarikalarKutusu: Removed both `_ne!` tests.

src/replacer.rs
HarikalarKutusu: I'm also eager to learn how brackets, punctuation and spaces are handled in 2500+ living languages :)

HarikalarKutusu:
@MichaelKohler, all is OK for now.  I hope this might help the lower-resourced Wikis as sentence count and the text-corpus quality in CV for all.

MichaelKohler (approved):
Thank you for working on this and sorry again it took so long to review.

PR #202, opened 2023-10-25
Adding regex replacement feature

raivisdejus:
Adding another replacement option, that can process regexs. This can be used to split longer sentences into smaller chunks.

In my tests for Latvian, this can yield ~25% more sentences in the final output.


MichaelKohler:
Thanks for submitting this! I will have a closer look at this PR tomorrow.

README.md
```diff
@@ -155,30 +155,31 @@ cargo run --release -- -l en -d ../texts/ extract-file >> file.en.txt
...
+| remove_brackets_list          | Removes (possibly nested) user defined brackets and content inside them `(anything [else])` from the sentence before replacements and checking other rules                      | Array of matching brackets: each configuration is an Array of two values: `["opening_bracket", "closing_bracket"]`. See example below. | []
+| replacements                  | Replaces abbreviations or other words according to configuration. This happens before any other rules are checked.                                                              | Array of replacement configurations: each configuration is an Array of two values: `["search", "replacement"]`. See example below.    | nothing gets replaced
+| regex_replacement_list        | Finds regex and makes replacements within found patterms. This happens before any other rules are checked.                                                                      | Array of configurations: each configuration is an Array of three values: `["regex", "search", "replacement"]`. See example below.     | nothing gets replaced
```
MichaelKohler:
```suggestion
| regex_replacement_list        | Finds regex and makes replacements within found patterns. This happens before any other rules are checked.                                                                      | Array of configurations: each configuration is an Array of three values: `["regex", "search", "replacement"]`. See example below.     | nothing gets replaced
```

README.md
```diff
@@ -239,6 +240,29 @@ Input: I am foo test a test
...
+  # Split glued sentences
+  ["\\ [a-z]{3,}\\.[A-Z][a-z]{2,}\\ ", ".", ". "],
+  
+  # Split long sentences
+  ["\\b(?:\\S+\\s+){15,}\\S+[.!?]", ", but ", ". But "],
+]
+```
+
+This will find words that glue two sentences and will add a space to un-glue them. 
+And will split a long sentence in two smaller.
```
MichaelKohler: I think this is a good example and easily understandable, thanks for this thorough documentation. In the context of Wikipedia extracts, more sentences might actually mean less content, as a sentence might be fulfilling all rule requirements, but then gets split into two. And then only one of them gets picked. Of course this heavily depends on how many potential sentences a given article has. In many cases (such as yours), this might be beneficial, but it doesn't always have to be. Might be worth it to write a short explanation here for that as well.

src/replacer.rs
```diff
@@ -28,6 +28,19 @@ pub fn replace_strings(rules: &Rules, raw: &str) -> String {
         }
     }
 
+    // regex replacements
+    for regex_replacement in rules.regex_replacement_list.iter() {
+        if Value::as_array(regex_replacement).unwrap().len() == 3 {
+            let regex = Regex::new(&regex_replacement[0].as_str().unwrap()).unwrap();
```
MichaelKohler:
```suggestion
            let regex = Regex::new(regex_replacement[0].as_str().unwrap()).unwrap();
```

src/replacer.rs
```diff
@@ -28,6 +28,19 @@ pub fn replace_strings(rules: &Rules, raw: &str) -> String {
         }
     }
 
+    // regex replacements
+    for regex_replacement in rules.regex_replacement_list.iter() {
+        if Value::as_array(regex_replacement).unwrap().len() == 3 {
```
MichaelKohler, line 33: This made me wonder if this implementation should go further than just with 3 values. Initially I thought such a regex implementation would only take two arguments and basically work like the `replace_all` function. But thinking about it, I can absolutely see why 3 arguments can be even more helpful, though many use cases could also be covered by named capture groups (but not all!).

Would you be interested in implementing a second form of this that accepts two arguments and replaces every matched occurrence with that string? This of course could be done outside this PR as a follow-up.

MichaelKohler (changes requested):
Thanks for submitting this PR! I have left a few comments, but nothing major. Please ping me once ready for another round of review and we can get this merged as soon as possible :)

README.md
HarikalarKutusu: I think this is where it is most worthwhile: If the article does not have enough sentences to select from (<3) because of the rules, especially max_words and/or max_characters. At that time, this algorithm can kick in and try to produce split sentences.

There is no way for us to know if pre-split or post-split can produce more "valuable" sentences. But many of the "sub-sentences" might be simple introductory wording etc.

MichaelKohler: I agree, and that's exactly why I would prefer a short sentence explaining that, so people don't just blindly copy. If we have an indication that it works in all corpuses, then we could also just do it by default.

raivisdejus: Added a note on sentence splitting.

